PR #50, opened 2014-12-15
add hh for hack to file types

jchpi:
fixing https://github.com/atom/atom/issues/4416

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -13,6 +13,7 @@
   'phpt'
   'phtml'
   'profile'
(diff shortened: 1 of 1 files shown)


izuzak:
@kevinsawicki Is this still relevant after https://github.com/atom/atom/issues/6847 and https://github.com/steelbrain/language-hack?

kevinsawicki:
Yeah, I think this can be closed out now that that package exists and will be chosen for the `.hh` extension.

PR #54, opened 2014-12-26
Properly identify finally keyword

AdamBuczynski:
This properly identifies "finally" as an exception keyword instead of a constant.


kevinsawicki:
Thanks for this

PR #62, opened 2015-02-19
Adding semicolons after control structures in .text.html.

robballou:
While not required in the scenario of html/php, ending semicolons make it easier to add new code within existing PHP tags. I had thought that some coding standards mandated this, but I didn't find a supporting case so there is that caveat. Thanks for taking a look.


kevinsawicki:
Cool, thanks!

PR #65, opened 2015-03-28
Adding spec/ and prioritized combined operators

chemoish:
Similar to https://github.com/atom/language-javascript/pull/119

Capture combined operators.

Ref: http://php.net/manual/en/language.operators.assignment.php


kevinsawicki:
Awesome, thanks for adding specs :+1:

PR #75, opened 2015-05-19
Add non word characters to php that excludes the dollar sign from a valid variable name

mostafaeweda:
@kevinsawicki @bolinfest

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "language-php",
-  "version": "0.24.0",
+  "version": "0.23.0",
   "description": "PHP language support in Atom",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "language-php",
-  "version": "0.23.0",
+  "version": "0.24.0",
```
kevinsawicki: This gets changed as part of `apm publish`, please revert this change.

mostafaeweda: Oh, OK

kevinsawicki:
Thanks!

PR #73, opened 2015-05-14
Fixed so namespaces syntax works on same lime as <?php

frozzare:
This pull request will fix https://github.com/atom/language-php/issues/29

Issue #63, opened 2015-02-27
Auto-indent issues - square brackets, function arguments, heredoc/nowdoc


crantok:
Is this the right place to ask about indenting issues that I am experiencing when editing PHP? Or is auto indenting actually handled by a different module?

The first two examples have very obvious work-arounds and are just annoying. The heredoc issue requires any heredoc strings to be revisited every time anyone auto-indents the whole file.

Indenting square-bracket array syntax over multiple lines only works with a trailing coma, so this works:

```php
class Foo
{
	function bar ()
	{
		return
		[
			[
				// stuff
			],
			[
				// more stuff
			],
		];
	}
}
```

But this does not:

```php

class Foo
{
	function bar ()
	{
		return
		[
			[
				// stuff
			],
			[
				// more stuff
				]
			];
		}
	}	
```

Indenting  a multi-line function call only works if the closing parenthesis is on the same line as the final parameter, so this works:

```php
class Foo
{
	function bar ()
	{
		foobar(
		extremely_long_and_tedious_variable_name_1,
		extremely_long_and_tedious_variable_name_2,
		extremely_long_and_tedious_variable_name_3 );
	}
}
```

But this does not:

```php
class Foo
{
	function bar ()
	{
		foobar(
		extremely_long_and_tedious_variable_name_1,
		extremely_long_and_tedious_variable_name_2,
		extremely_long_and_tedious_variable_name_3
	);
}
}
```

And finally, heredoc and nowdoc syntax are both broken by auto-indent:

```php
class Foo
{
	function bar ()
	{
		return
		<<<'HTML'
		<blockquote>Somthing pithy</blockquote>
		HTML
	}
}
```

JesseLeite:
@crantok Wonder if your issue is related to this PR? https://github.com/atom/language-php/pull/96

crantok:
@JesseLeite Yes, that definitely relates to at least one of the cases above.

I stopped using atom in favour of lighttable soon after reporting this issue. LT uses CodeMirror for syntax highlighting and indenting. It is very easy to sanity-check issues using the [codemirror online demos](https://codemirror.net/mode/) and all formatting issues are concentrated in one project. I'm not saying that's the reason I swapped, just that it seems to provide a highly favourable workflow for dealing with formatting issues.

charlesqwu:
[...]

Issue #57, opened 2015-01-28
HEREDOC highlighting broken


wyqydsyq:
I have the following code:

``` php
$this->data['text_message'] .= <<<TRACKER
        <!-- Facebook Conversion Code for Checkout Success -->
<script>(function() {
var _fbq = window._fbq || (window._fbq = []);
if (!_fbq.loaded) {
var fbds = document.createElement('script');
fbds.async = true;
fbds.src ='//connect.facebook.net/en_US/fbds.js';
var s =document.getElementsByTagName('script')[0];
s.parentNode.insertBefore(fbds, s);
_fbq.loaded = true;
}
})();
window._fbq = window._fbq || [];
window._fbq.push(['track', '',{'value':'0.00','currency':'AUD'}]);
</script>
<noscript><img height="1" width="1"alt="" style="display:none" src="https://www.facebook.com/tr?ev=&amp;cd[value]=0.00&amp;cd[currency]=AUD&amp;noscript=1"/></noscript> 
TRACKER;
```

But atom doesn't highlight the contents of the HEREDOC as a string:

![Example](http://i.imgur.com/2F10GqA.png)

The first `<` is correctly given the appropriate classes (`class="string unquoted heredoc php"`) but the class doesn't persist to any of following tokens:
![Example 2](http://i.imgur.com/Jd3Nnyw.png)

50Wliu:
Refs #10.

pyrokinetiq:
It seems in #10 they're getting the correct behaviour of the heredoc being treated as string until a PHP tag is encountered, my case is only the first character of the heredoc is treated as a string regardless of the heredoc's contents.

halleycarleton:
+1
![HERDOC Syntax Bug]
(https://i.gyazo.com/4dff4b2bb513ca92bae6bc17a4a31f61.png)

*Windows 8.1 / Atom 0.179.0*
*Ubuntu 14.04 / Atom 0.174.0*

Talon1024:
I see that regular expressions are being used to highlight syntax elements like this. Here's a regular expression that matches a PHP Heredoc:
![PHP Heredoc](https://cloud.githubusercontent.com/assets/6550500/7689837/4ec61e2a-fd7d-11e4-9c6e-1d85964a80bc.png)
Here's a regular expression that matches a PHP Nowdoc:
[...]

Issue #10, opened 2014-03-03
Heredoc containing a closing PHP tag breaks syntax highlighting


jaikdean:
I have a file containing the following:

``` php
        $this->xml = <<<EOD
<?xml version="1.0"?>
<methodCall>
<methodName>{$this->method}</methodName>
<params>

EOD;
```

The syntax formatting is picking up the `?>` within the heredoc as a closing PHP tag, and as such the syntax highlighting is broken from this point on in the file.

![screen shot 2014-03-03 at 10 58 40](https://f.cloud.github.com/assets/348408/2308690/d8630ad8-a2c2-11e3-9fdb-bd1544bb38bc.png)

nicksayre:
Syntax highlighting also breaks with inclusion of a single-quote in a heredoc.

![screen shot 2015-09-16 at 10 08 12 am](https://cloud.githubusercontent.com/assets/1807967/9908774/c575671a-5c5a-11e5-981d-7a0ffd4d4b58.png)

TheJosh:
Likewise for single quotes (`'`) or double quotes (`"`) within a heredoc

zr9:
There is 2 problems in atom/core probably.
Because grammar ignore "repository" and "injection" rules.

Good news so i'm can fix it by rewriting without that rules, here is simple trial example
![atom_issue_3](https://cloud.githubusercontent.com/assets/13107923/12418160/6fce551c-bebe-11e5-8d6e-070b61e692cb.png)

But i'm need someone to help me write bug about grammar(En is not my native, so i'm sometimes unclear as hell).

And as well i'm need someone who can explain tricky parts in itself heredoc syntax,
like scope "string.unquoted.heredoc.php" now contain in rule **double quote** 
and scope "string.unquoted.heredoc.nowdoc.php" now contain **single quote**
so i'm need to know diff and probably some examples of different languages(Js,SQL,Json,Regex,CSS) heredoc to test

Ingramz:
@zr9 we've already concluded that it would be worthwhile fixing the root cause of this problem in first-mate package ([see related discussion](https://github.com/atom/language-php/pull/77)), so that all other bugs would fix on their own. The current solution is rather elegant (although, not working).

But it's still nice seeing someone giving their go at it.

zr9:
@Ingramz Oh i'm not read references, so core bug already submitted. Then np :) But current solution is seems like working, so it can be fine as temporally fix until atom fixed "repository"

Ingramz:
[...]

PR #77, opened 2015-05-24
Don't accept spaces after heredoc beginning identifier. (was: Make Heredoc syntax highlighting work)

Talon1024:
I've modified some of the regular expressions so that Heredoc syntax highlighting now works.

Also, I've disabled the "internal heredocs" because, as far as I know, there is no such thing in PHP.

### DIFF
--- a/spec/php-spec.coffee
+++ b/spec/php-spec.coffee
@@ -437,3 +437,151 @@ describe 'PHP grammar', ->
         expect(tokens[1][9]).toEqual value: '{', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'punctuation.section.scope.begin.php']
         expect(tokens[1][10]).toEqual value: '}', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'punctuation.section.scope.end.php']
         expect(tokens[1][11]).toEqual value: ';', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'punctuation.terminator.expression.php']
+
+      it 'should tokenize a simple heredoc correctly', ->
+        tokens = grammar.tokenizeLines """<?php
+        $a = <<<HEREDOC
+        I am a heredoc
+        HEREDOC;"""
+
+        expect(tokens[1][0]).toEqual value: '$', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'variable.other.php', 'punctuation.definition.variable.php']
+        expect(tokens[1][1]).toEqual value: 'a', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'variable.other.php']
+        expect(tokens[1][2]).toEqual value: ' ', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php']
+        expect(tokens[1][3]).toEqual value: '=', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'keyword.operator.assignment.php']
+        expect(tokens[1][4]).toEqual value: ' ', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php']
+        expect(tokens[1][5]).toEqual value: '<<<', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.unquoted.heredoc.php', 'punctuation.definition.string.php']
+        expect(tokens[1][6]).toEqual value: 'HEREDOC', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.unquoted.heredoc.php', 'keyword.operator.heredoc.php']
+        expect(tokens[2][0]).toEqual value: 'I am a heredoc', scopes ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.unquoted.heredoc.php']
+        expect(tokens[3][0]).toEqual value: 'HEREDOC', scopes ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.unquoted.heredoc.php', 'keyword.operator.heredoc.php']
+        expect(tokens[3][1]).toEqual value: ';', scopes ['text.html.php', 'meta.embedded.block.php', 'source.php', 'punctuation.terminator.expression.php']
+
+      it 'should tokenize a longer heredoc correctly', ->
+        tokens = grammar.tokenizeLines """<?php
+        $a = <<<GITHUB
+        This is a plain string.
+        SELECT * FROM github WHERE octocat = 'awesome' and ID = 1;
+        <strong>rainbows</strong>
+
+        if(awesome) {
+            doSomething(10, function(x){
+              console.log(x*x);
+            });
+        }
+        GITHUB;"""
+
+        expect(tokens[1][0]).toEqual value: '$', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'variable.other.php', 'punctuation.definition.variable.php']
+        expect(tokens[1][1]).toEqual value: 'a', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'variable.other.php']
+        expect(tokens[1][2]).toEqual value: ' ', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php']
+        expect(tokens[1][3]).toEqual value: '=', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'keyword.operator.assignment.php']
+        expect(tokens[1][4]).toEqual value: ' ', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php']
+        expect(tokens[1][5]).toEqual value: '<<<', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.unquoted.heredoc.php', 'punctuation.definition.string.php']
+        expect(tokens[1][6]).toEqual value: 'GITHUB', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.unquoted.heredoc.php', 'keyword.operator.heredoc.php']
+        expect(tokens[2][0]).toEqual value: 'This is a plain string.', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.unquoted.heredoc.php']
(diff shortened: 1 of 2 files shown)


Ingramz:
heredoc_interior rule is supposed to provide a semantic way of embedding different sources. While not specifically a language feature, it makes an useful convention to follow and the highlighting is there to help with that.

This works with linguist and textmate's PHP bundle in its current form.

```php
<?php

$a = <<<GITHUB
This is a plain string.
SELECT * FROM github WHERE octocat = 'awesome' and ID = 1;
<strong>rainbows</strong>

if(awesome) {
    doSomething(10, function(x){
      console.log(x*x);
    });
}
GITHUB;

$b = <<<SQL
SELECT * FROM github WHERE octocat = 'awesome' and ID = 1;
SQL;

$c = <<<HTML
<strong>rainbows</strong>
HTML;

$d = <<<JAVASCRIPT
if(awesome) {
    doSomething(10, function(x){
      console.log(x*x);
    });
}
JAVASCRIPT;

$e = <<<JSON
{"menu": {
  "id": "file",
  "value": "File",
  "popup": {
    "menuitem": [
      {"value": "New", "onclick": "CreateNewDoc()"},
      {"value": "Open", "onclick": "OpenDoc()"},
      {"value": "Close", "onclick": "CloseDoc()"}
    ]
  }
}}
JSON;

?>
```

As for the actual issue, I think it has something to do with the lookahead + `(?!\\G)` plainly not working in this case (possibly an issue with first-mate and not with the grammar itself), but right now I can't really comment on that.

Talon1024:
Why were lookahead and lookbehind (AKA lookaround) being used in the first place? In my experience, lookaround does not work well with whitespace.

Ingramz:
The main reason is because textmate grammar had it - why change something that has been proven to work, right?

The actual reason however is because if you match something without lookaround, you cannot match it again. Lookaheads in this case allow to match a generic version of the heredoc/nowdow beginning and then more specific versions can be matched instead inside the block, a nifty way of enabling the functionality for heredoc_interior here. As far as I know the whitespace is no problem. 

What confuses me a little the usage of `(?!\\G)` pattern in combination with the lookahead.

Unfortunately I cannot look into this issue before Saturday.

Talon1024:
Now that I think of it, the point of doing a lookahead for the beginning of the heredoc is to match the position before the heredoc.

I guess the "ending" token should then match the position after the ending identifier.

Ingramz:
Either way for some odd reason right now only the first character from `<<<SOMETHING` is in the heredoc scope and not all of the block.

Pinging @kevinsawicki / @nathansobo for some help with demystifying first-mate behavior in this case.

Talon1024:
I think the pattern `(?!\\G)` will match any position/character that isn't proceded by the beginning of the start of the match. The beginning and end regular expressions are, of course, evaluated in that order, and the ending regular expression can only be evaluated once. Since the first `<` sign isn't proceded by the start of the match (it appears after the start of the match), it is the only character in the heredoc scope.

Ingramz:
Hmm... after sleeping on it, you are right.

Ingramz:
Still I do not believe that changing the grammar is the solution here. We are talking of at least 3 implementations where this exact regex works and now it just doesn't.

nathansobo:
I probably broke this with some changes to the parser intended to save memory. I will look into it Monday at the latest.

Ingramz:
@nathansobo it's possible that heredoc strings have never worked in Atom.

Talon1024:
[Heredoc syntax highlighting has been broken since Atom 0.174](https://github.com/atom/language-php/issues/57).

Ingramz:
@Talon1024 Thanks, my bad.

Ingramz:
@Talon1024 0.174 did not break it, so the search for a working version continues.

Talon1024:
I just tried Atom 0.45, and heredoc syntax highlighting didn't work in that version either. On the other hand, maybe try matching as much as possible in between the beginning and end tokens instead of as little as possible... Just a thought.

Talon1024:
So, I've compiled my own build of TextMate using the source code from their GitHub repo, and I've played around with the TextMate PHP bundle using the built-in bundle editor.

It works properly if no changes are made to the bundle.

If I remove the `#heredoc_interior` include from the TextMate grammar, the highlighting breaks, and the code ends up being highlighted in the same way it is in Atom now (only the first < character is in the heredoc scope)

It is partially fixed, with no syntax highlighting for SQL, HTML, JS, or JSON if I make the equivalent changes that I made to the TextMate bundle.

These findings have led me to believe that "includes" in TextMate work differently than they do in Atom. I think the way it works in TextMate is that, if a match for the included grammar element is found at the starting position of the parent grammar element, it will highlight from that point using the included grammar element regexes.

However, [The grammar for this package has been changed in the past](#73), and it's ultimately up to whoever is in charge of Atom and this bundle if they want 1:1 TextMate bundle compatibility.

50Wliu:
This no longer merges cleanly; @Talon1024 would you mind rebasing?

Talon1024:
OK, I've changed my repo to be up to date with this repo, and I redid my changes from scratch.

50Wliu:
I'll try to review this tonight.

Ingramz:
Not that I'd like to see this to be merged, I believe that this workaround won't motivate to solve the underlying cause in first-mate or underlying regular expression engine. I hope that merging this right now won't do that.

Also someone should open an issue in first-mate for that (I can do it later tonight).

Talon1024:
Oh my goodness. I was able to fix this without changing any of the expressions, but simply moving the `heredoc_interior` rule out of a nested `repository`.

Take a look at the structure of the JSON generated from php.cson:
phpgrammarjsonstructure

Notice the `heredoc_interior` rule is in a nested `repository`? If I move it out of the nested `repository`, heredocs will highlight correctly.
phpgrammarhackfixed

So, the problem is really first-mate ignoring nested `repositories`.

Ingramz:
This seems awfully familiar when I tried to fix it. I ended up posting this: https://github.com/atom/first-mate/issues/38#issuecomment-138676408

Talon1024:
In this commit, I've restored the original structure of php.cson, but I've modified the expressions so that they won't accept spaces after the heredoc beginning identifier, since that will cause a PHP syntax error.

Ingramz:
LGTM :+1:

50Wliu:
Please add tests to this :smile:.

Ingramz:
Also note that we don't want to highlight newline characters incorrectly, which is what capture group `(\s*) ` does, although it's fine for the case when you aren't highlighting those characters. That is why I suggested  `([ \t]*)` instead. Atom doesn't have highlighting support for end-of-lines right now (although it still matches them), but in order to keep the grammar accurate, I suggest you updating all the corresponding capturing groups to match space and tab only.

Talon1024:
Keep in mind that `$` is an anchor that matches the end of the line, so I don't think (\s*) will match a newline if it comes before a `$`, which it does.

Talon1024:
I just added some breaking tests and modified the grammar so that a zero-length string at the end of the line will not get tokenized as invalid/illegal.

Ener-Getick:
@Talon1024 did you stop working on this ?
It would be amazing to this working and merged !

Aldlevine:
Any progress on this? It would be an awesome feature to have! :+1:

Wardevour:
I have a block like:
```php
$tst = <<<HTML
<script type="text/javascript">
try {
    // code goes here...
}
catch(e) {
    alert(e.message);
}
</script>
HTML;

```

JS in a heredoc is probably frowned upon, but the syntax highlighting stops working on line 6 because it's not a correct PHP catch block (missing Exception class type). Will this PR fix my issue or am I having a different problem?

UziTech:
> Notice the heredoc_interior rule is in a nested repository? If I move it out of the nested repository, heredocs will highlight correctly.

If there was a fix for this over a year ago why hasn't this been fixed yet? I realize it would only be a temporary but it sure would've beaten no fix while we wait on an issue that some other package [isn't even working on.](https://github.com/atom/first-mate/issues/38#issuecomment-141123296)

Can we please just get a temporary fix done while we wait on this pull request?

50Wliu:
I would much rather the underlying issue be fixed than piling on "temporary" workarounds, because from experience it then becomes "oh, there's a workaround, so we can deprioritize fixing the actual issue" and the workaround becomes not-so-temporary.

At any rate, I can't merge this even if I wanted to because it has conflicts.

UziTech:
If it is between no fix for a year+ (and god knows how much longer) or a temporary fix, I would rather have the temporary fix. Especially since [this](https://github.com/atom/language-php/issues/10) [obviously](https://github.com/atom/language-php/issues/57) [affects](https://github.com/atom/language-php/issues/63) [a](https://github.com/atom/language-php/issues/78) [lot](https://github.com/atom/language-php/issues/113) [of](https://github.com/atom/language-php/issues/133) [people](https://github.com/atom/language-php/issues/155) 

If the underlying issue was a priority, I would agree with you. But it is obviously not.

I created a new pull request #184

spec/php-spec.coffee
```diff
@@ -437,3 +437,151 @@ describe 'PHP grammar', ->
...
+        expect(tokens[1][3]).toEqual value: '=', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'keyword.operator.assignment.php']
+        expect(tokens[1][4]).toEqual value: ' ', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php']
+        expect(tokens[1][5]).toEqual value: '<<<', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.unquoted.heredoc.php', 'punctuation.definition.string.php']
+        expect(tokens[1][6]).toEqual value: 'HEREDOC', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.unquoted.heredoc.php', 'keyword.operator.heredoc.php']
+        expect(tokens[2][0]).toEqual value: 'I am a heredoc', scopes ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.unquoted.heredoc.php']
+        expect(tokens[3][0]).toEqual value: 'HEREDOC', scopes ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.unquoted.heredoc.php', 'keyword.operator.heredoc.php']
+        expect(tokens[3][1]).toEqual value: ';', scopes ['text.html.php', 'meta.embedded.block.php', 'source.php', 'punctuation.terminator.expression.php']
```
UziTech: lines 454 - 456

There should be a colon (:) after `scopes`

50Wliu:
Superseded by #184.

PR #79, opened 2015-05-30
Sync with TextMate's PHP bundle

Ingramz:
It looks like TextMate's bundle has gotten some love during the time this grammar was converted to Atom. All changes since textmate/php.tmbundle@7bcc57b have been applied to this grammar.

Incidentally this reverts changes from v0.23.0, but does not break anything (I hope) as the bundle had the same issue solved already.


Ingramz:
One thing needs to be sorted first: there is a duplicate for `trait` keyword now, will have to look into this before merging.

I'll probably make another commit to resolve any duplicates.

josegonzalez:
Bump :)

50Wliu:
@Ingramz can you by any chance add specs for the new changes?

Ingramz:
@50Wliu I'll try adding some in a day or two.

Ingramz:
Is there anything else that needs to be done?

50Wliu:
:articulated_lorry: 

(Also, for next time: we don't notified by new commits, so after a commit, it would be great to just ping us again :bell:)

Ingramz:
That I figured, thanks for confirming!

PR #83, opened 2015-06-22
Added syntax support for PHPDoc api tag

frozzare:
Added syntax support for `@api` 

See http://phpdoc.org/docs/latest/references/phpdoc/tags/api.html or https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#81-api


50Wliu:
Thanks for this!

PR #84, opened 2015-06-22
Fixed open tag completion

### LINKED ISSUE #76: Open tag completion causes wrong tag <?<?php
Atom Version 0.201.0 (0.201.0) on Mac OS X

I have two things I run into when opening php tags:

**Unwanted completion**
When typing:

```
<?php
```

I'd expect to be able to press enter and go on typing, instead I have to press escape first to turn off the autocomplete window. If I don't press esc first I end up having:

```
<?<?php  ?>
```

**Wrong completion**
When typing the following:

```
<?p
```

The autocomplete window pops up as expected, when pressing enter and triggering the completion it replaces with:

```
<?<?php  ?>
```

I'd expect to get:

```
<?php ?>
```

Or maybe even better when it is on the first line of the file

```
<?php
```
This due classes mostly not having a closing PHP tag at the end of the file.

frozzare:
Created a new pull request from a branch and not master.

This will fix #76

### DIFF
--- a/snippets/language-php.cson
+++ b/snippets/language-php.cson
@@ -2,9 +2,9 @@
   '$GLOBALS[\'…\']':
     'prefix': 'globals'
     'body': '$GLOBALS[\'${1:variable}\']${2: = }${3:something}${4:;}$0'
-  '?>…<?php':
+  '<?php … ?>':
     'prefix': 'php'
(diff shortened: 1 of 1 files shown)


kevinsawicki:
Awesome, nice fix :+1:

PR #85, opened 2015-06-22
Added support for phpdoc method, property and source tags

frozzare:
Noted that this PHPDoc tags was missing from the syntax highlighting. This pull request will add support for the missing tags.

Reference: http://phpdoc.org/docs/latest/references/phpdoc/tags/

* `@method`
* `@property`
* `@property-read`
* `@property-write`
* `@source`


kevinsawicki:
Thanks for adding these :shipit:

PR #103, opened 2015-09-24
Improve "function __construct" snippet

henriquemoody:
- Create new line after brackets (as PSR-2 recomments)
- Add visibility to the method (`public` as default)
- Add type to the first argument
- Define a property name as the same of the first argument

***
The snippet should create something like:
```php
public function __construct(Type $foo = null)
{
    $this->foo = $foo;
}
```

### DIFF
--- a/snippets/language-php.cson
+++ b/snippets/language-php.cson
@@ -7,7 +7,7 @@
     'body': 'php $0 ?>'
   'function __construct':
     'prefix': 'con'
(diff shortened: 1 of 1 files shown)


Ingramz:
LGTM! :+1:

Behaves a lot better than the previous one.

henriquemoody:
It's been a long time since I've opened this pull request, is it going to be merged?

kevinsawicki:
Great, thanks for this, apologies for the delay

henriquemoody:
No problem :)

PR #106, opened 2015-10-22
Generate Grammar using PHP Manual as input

klorenz:
... and use atom-syntax-tools, to make grammar easier to maintain.

Intention for this was a fix for indented phpdoc comments, which were not highlighted correctly.  Fixing the grammar was almost impossible, because it is huge and has lines of several thousand characters long.  So I started to move the grammar beeing generated using atom-syntax-tools and using the PHP Documentation as input.

There is still a lot to be done, but this state fixes the initial indented phpdoc bug and the grammar
is a lot easier to maintain.  Before I continue with removing unneeded noise from php-grammar.coffee,
I want to see, if you will accept such big changes.

Best regards, Kiwi


joehoyle:
This looks pretty good to me, what's the next step here on this? Anyone from the Atom team able to review this?

PR #110, opened 2015-11-07
Echo and print are language control keywords not functions

thisispiers:


50Wliu:
Please add specs to this.

MaximSokolov:
I disagree. `echo` and `print` are `support.function`. See [naming conventions](http://manual.macromates.com/en/language_grammars#naming_conventions)

thisispiers:
Straight from the [PHP docs](http://php.net/manual/en/function.echo.php])
> echo is not actually a function (it is a language construct), so you are not required to use parentheses with it.

I just wanted my syntax highlighter to show a different colour

MaximSokolov:
You can add something like this to your stylesheet:
```less
atom-text-editor::shadow {
  .support.function.construct.php {
    color: red;
  }
}
```

thisispiers:
Thanks, but that colours anything else with `.support.functions.construct.php`
I wanted `echo` to be the same colour as `foreach` and `if`

MaximSokolov:
Could you change
```coffee
'name': 'support.function.construct.php'
```
to 
```coffee
'name': 'support.function.construct.$.php'
```
It would be `support.function.construct.(echo|print)`, so you will be able to change color for them.

thisispiers:
I'm sure on how to spec it but I suppose that works for more people :)

Ingramz:
`echo` and `print` are definitely not control keywords (see the naming conventions above). Per [PHP spec](http://php.net/manual/en/reserved.keywords.php) they aren't actually functions either.

At best these two can be described as keywords which behave like functions, but they are not functions. Notice the use of the term language **constructs**, which is probably why these not-functions were named this way originally.

Also consider that there are themes which directly target `support.function.construct` or `support.function.construct` just so that `echo` and `print` would have distinct coloring in PHP (and in some other places), you'd break a lot of themes by just doing so.

If someone wanted to follow PHP spec and create a semantically named scope name from the TextMate conventions, `keyword.other.something` would be the most appropriate scope name for it in my opinion (again, this is arguable, I don't like it falling to 'other' and I don't advise to change it anyway).

As @MaximSokolov mentioned, if you wanted to change something for your taste in terms of looks, consider modifying your existing theme or apply the change using custom stylesheet. It shouldn't be difficult to find out what color your keywords are in your current theme.

thisispiers:
CSS doesn't know about the text content of an element so there was no other way of distinguishing between ordinary functions and these two non-control-non-functions than to mess with the language interpreter.

Would people object to `support.function.construct.output` for instance? This at least separates these two without using existing class names?

Ingramz:
Extending the scope name would be fine for that case.

MaximSokolov:
> `support.function.construct.output`

:+1:

spec/php-spec.coffee
```diff
@@ -435,3 +435,13 @@ describe 'PHP grammar', ->
...
         expect(tokens[1][10]).toEqual value: '}', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'punctuation.section.scope.end.php']
         expect(tokens[1][11]).toEqual value: ';', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'punctuation.terminator.expression.php']
+        
+      it 'should tokenize non-function-non-control operations correctly', ->
+        tokens = grammar.tokenizeLines "<?php\necho 'test';"
+
+        expect(tokens[1][0]).toEqual value: 'echo', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'support.function.construct.output.php']
+        expect(tokens[1][1]).toEqual value: ' ', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php']
+        expect(tokens[1][2]).toEqual value: '\'', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.quoted.single.php', 'punctuation.definition.string.begin.php']
+        expect(tokens[1][3]).toEqual value: ' test', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.quoted.single.php', 'meta.string-contents.quoted.single.php']
```
MaximSokolov: ' test' -> 'test'

50Wliu:
Glad we resolved the naming issue :)

PR #114, opened 2015-12-20
Fix assignment operator consuming comparison operators

denis-sokolov:
`===` is currently being treated as three assignment operators.

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -1189,8 +1189,16 @@
         'name': 'keyword.operator.error-control.php'
       }
       {
+        'match': '(===|==|!==|!=|<>)'
+        'name': 'keyword.operator.comparison.php'
+      }
+      {
         'match': '=|\\+=|\\-=|\\*=|/=|%=|&=|\\|=|\\^=|<<=|>>='
         'name': 'keyword.operator.assignment.php'
+      }
+      {
+        'match': '(<=|>=|<|>)'
+        'name': 'keyword.operator.comparison.php'
       }
       {
         'match': '(\\-\\-|\\+\\+)'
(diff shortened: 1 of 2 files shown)


grammars/php.cson
```diff
@@ -1216,6 +1212,10 @@
         'name': 'keyword.operator.comparison.php'
       }
       {
+        'match': '=|\\+=|\\-=|\\*=|/=|%=|&=|\\|=|\\^=|<<=|>>='
```
50Wliu: While you're at it, it looks like `=` should be matched last in this pattern.

50Wliu:
Looks like some specs broke - could you please take a look at that?  Thanks!

denis-sokolov:
I did not know that you had a spec, since `npm test` did diddly squat! :)
Now that I know, I’ve added a spec for the PR and all specs are passing.

grammars/php.cson
```diff
@@ -1189,10 +1189,18 @@
...
+        'match': '(===|==|!==|!=|<>)'
+        'name': 'keyword.operator.comparison.php'
+      }
+      {
         'match': '=|\\+=|\\-=|\\*=|/=|%=|&=|\\|=|\\^=|<<=|>>='
         'name': 'keyword.operator.assignment.php'
       }
       {
+        'match': '(<=|>=|<|>)'
+        'name': 'keyword.operator.comparison.php'
```
50Wliu: Why can't the two comparison matches be combined into one?

50Wliu:
> I did not know that you had a spec, since npm test did diddly squat! :)

Try `apm test` next time :wink:.

grammars/php.cson
denis-sokolov: Because `<<=` needs to match before `<=`, but `===` needs to match before `=`.
If there's a way to do that with no two comparison matches, great.

50Wliu: Hmm, that *is* tricky.  Let's just leave it as-is then :).

50Wliu:
One last question before I merge: is `!==` getting tokenized correctly?  It looks like it's being overridden by `==`.

denis-sokolov:
Added a spec for that - apparently, no. I am not exactly sure, why.

50Wliu:
:ship:

PR #123, opened 2016-01-24
Cleaning: removed garbage symbols from regexes

zr9:
removed garbage symbols from regexes, and make them more nice look

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -164,7 +164,11 @@
         'captures':
           '0':
             'name': 'punctuation.definition.comment.php'
-        'comment': 'This now only highlights a docblock if the first line contains only /**\n\t\t\t\t\t\t\t\t- this is to stop highlighting everything as invalid when people do comment banners with /******** ...\n\t\t\t\t\t\t\t\t- Now matches /**#@+ too - used for docblock templates: http://manual.phpdoc.org/HTMLframesConverter/default/phpDocumentor/tutorial_phpDocumentor.howto.pkg.html#basics.docblocktemplate'
+        'comment': '''
+          This now only highlights a docblock if the first line contains only /**
+          - this is to stop highlighting everything as invalid when people do comment banners with /******** ...
+          - Now matches /**#@+ too - used for docblock templates: http://manual.phpdoc.org/HTMLframesConverter/default/phpDocumentor/tutorial_phpDocumentor.howto.pkg.html#basics.docblocktemplate
+        '''
         'end': '\\*/'
         'name': 'comment.block.documentation.phpdoc.php'
         'patterns': [
@@ -219,7 +223,17 @@
   'constants':
     'patterns': [
       {
-        'begin': '(?xi)(?=\n\t\t\t            (\n\t\t\t                (\\\\[a-z_][a-z_0-9]*\\\\[a-z_][a-z_0-9\\\\]*)|\n\t\t\t                ([a-z_][a-z_0-9]*\\\\[a-z_][a-z_0-9\\\\]*)\n\t\t\t            )\n\t\t\t            [^a-z_0-9\\\\])'
+        'begin': '''
+          (?xi)
+          (?=
+            (
+              (\\\\[a-z_][a-z_0-9]*\\\\[a-z_][a-z_0-9\\\\]*)
+              |
+              ([a-z_][a-z_0-9]*\\\\[a-z_][a-z_0-9\\\\]*)
+            )
+            [^a-z_0-9\\\\]
+          )
+        '''
         'end': '(?i)([a-z_][a-z_0-9]*)?(?=[^a-z0-9_\\\\])'
         'endCaptures':
           '1':
@@ -267,7 +281,11 @@
             'name': 'support.constant.parser-token.php'
           }
           {
-            'comment': 'In PHP, any identifier which is not a variable is taken to be a constant.\n        \t\t\t\tHowever, if there is no constant defined with the given name then a notice\n        \t\t\t\tis generated and the constant is assumed to have the value of its name.'
+            'comment': '''
+              In PHP, any identifier which is not a variable is taken to be a constant.
+              However, if there is no constant defined with the given name then a notice
+              is generated and the constant is assumed to have the value of its name.
+            '''
             'match': '[a-zA-Z_\\x{7f}-\\x{ff}][a-zA-Z0-9_\\x{7f}-\\x{ff}]*'
             'name': 'constant.other.php'
           }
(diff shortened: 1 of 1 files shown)


grammars/php.cson
```diff
@@ -219,7 +223,17 @@
   'constants':
     'patterns': [
       {
-        'begin': '(?xi)(?=\n\t\t\t            (\n\t\t\t                (\\\\[a-z_][a-z_0-9]*\\\\[a-z_][a-z_0-9\\\\]*)|\n\t\t\t                ([a-z_][a-z_0-9]*\\\\[a-z_][a-z_0-9\\\\]*)\n\t\t\t            )\n\t\t\t            [^a-z_0-9\\\\])'
+        'begin': '''
+          (?xi)
```
50Wliu: Can you move this `(?xi)` to the above line?  So `'begin': '''(?xi)`

grammars/php.cson
```diff
@@ -219,7 +223,17 @@
...
+          (?=
+            (
+              (\\\\[a-z_][a-z_0-9]*\\\\[a-z_][a-z_0-9\\\\]*)
```
50Wliu: Comments describing what these lines do would be great.

grammars/php.cson
```diff
@@ -340,7 +365,26 @@
...
           '10':
             'name': 'invalid.illegal.non-null-typehinted.php'
-        'match': '(?xi)\n\t\t\t\t\t\t\t\\s*(array|callable) # Typehint\n\t\t\t\t\t\t\t\\s*(&)? \t\t\t\t\t# Reference\n\t\t\t\t\t\t\t\\s*((\\$+)[a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*) # The variable name\n\t\t\t\t\t\t\t(?:\n\t\t\t\t\t\t\t\t\\s*(?:(=)\\s*(?:(null)|(\\[)((?>[^\\[\\]]+|\\[\\g<8>\\])*)(\\])|((?:\\S*?\\(\\))|(?:\\S*?))))\t# A default value\n\t\t\t\t\t\t\t)?\n\t\t\t\t\t\t\t\\s*(?=,|\\)|\/[\/*]|\\#|$) # A closing parentheses (end of argument list) or a comma or a comment\n\t\t\t\t\t\t\t'
+        'match': '''
+          (?xi)
+          \\s*(array|callable)  # Typehint
+          \\s*(&)?  # Reference
+          \\s*(
+            (\\$+)[a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*
+          ) # The variable name
```
50Wliu: I think you meant for this comment to be on the above line?

grammars/php.cson
```diff
@@ -847,7 +907,16 @@
           }
           {
             'begin': '(?i)\\s*(?=[a-z_0-9\\\\])'
-            'end': '(?xi)(?:\n    \t\t\t                (?:\\s*(as)\\b\\s*([a-z_0-9]*)\\s*(?=,|;|$))\n    \t\t\t                |(?=,|;|$)\n    \t\t\t            )'
+            'end': '''
+              (?xi)
+              (?:
+                (?:
+                  \\s*(as)\\b\\s*([a-z_0-9]*)\\s*(?=,|;|$)
```
50Wliu: This can be put in one line...no need to separate it.
```(?:\\s*(as)\\b\\s*([a-z_0-9]*)\\s*(?=,|;|$))```

File: grammars/php.cson [hunk 5]
```diff
@@ -1062,7 +1131,74 @@
...
       }
       {
-        'begin': '(?x)\\s*\n\t\t\t\t\t    ((?:(?:final|abstract|public|private|protected|static)\\s+)*)\n\t\t\t\t        (function)\n\t\t\t\t        (?:\\s+|(\\s*&\\s*))\n\t\t\t\t        (?:\n\t\t\t\t            (__(?:call|construct|debugInfo|destruct|get|set|isset|unset|tostring|clone|set_state|sleep|wakeup|autoload|invoke|callStatic))\n\t\t\t\t            |([a-zA-Z0-9_]+)\n\t\t\t\t        )\n\t\t\t\t        \\s*\n\t\t\t\t        (\\()'
+        'begin': '''
+          (?x)
+          \\s*
+          (
+            (?:
+              (?:
+                final
```
50Wliu: These don't need to all be on separate lines.

grammars/php.cson
```diff
@@ -1062,7 +1131,74 @@
...
+                |
+                protected
+                |
+                static
+              )\\s+
+            )*
+          )
+          (function)
+          (?:
+            \\s+
```
50Wliu: These are short enough that you can combine them onto one line.

grammars/php.cson
```diff
@@ -1062,7 +1131,74 @@
...
+            |
+            (\\s*&\\s*)
+          )
+          (?:
+            (
+              __(?:
+                call
```
50Wliu: Same here - no need to separate into different lines (maybe two at most)

grammars/php.cson
```diff
@@ -2203,7 +2405,13 @@
           '4':
             'name': 'punctuation.definition.variable.php'
         'comment': 'Simple syntax with braces: "foo${bar}baz"'
-        'match': '(?x)\n\t\t\t\t\t\t((\\$\\{)(?<name>[a-zA-Z_\\x{7f}-\\x{ff}][a-zA-Z0-9_\\x{7f}-\\x{ff}]*)(\\}))\n\t\t\t\t\t\t'
+        'match': '''
+          (?x)
+          (
+            (\\$\\{)
```
50Wliu: The curly brackets here and on the next line don't need to be escaped.

50Wliu:
In general, please move the `(?x)` modifier on the same line as `begin` to save some space.

grammars/php.cson
zr9: Yup possible

grammars/php.cson
zr9: Possible, but take a lot of time and mostly will be pointless. Itself capturing group is explained in captures section as of parts for regex it possible to leave comments like "namespace" but it simply not improve regex understanding.

grammars/php.cson
zr9: Can be in above or there, because itself capture group is variable name, as well as regex inside is variable name.

grammars/php.cson
zr9: But is that way it not easier to read and modify?

grammars/php.cson
zr9: It for easy read/recognize. When it ass \\{ it become escaped character so easy to eye catch and see so it symbol not repetition.

PR #134, opened 2016-05-07
update constant.character.escape.php

### LINKED ISSUE #124: Broken syntax highlighting on regex with simple quotes
The highligtinh brokes when a regex with this form is simple quoted and since the regex until the end of file is hihglighted in red.
With double quotes it works.

bug-php-syntax

The first time I saw was on 1.0.19 version, now I'm using 1.3.3 and still ocurring.

bj7:
Fixes #124: Update regex-single-quoted to place the constant.character.escape.php inside the patterns array so as to fix the broken regex highlighting in single quoted php regular expressions. My screenshots and description are attached to the issue. Please let me know if I need to update, or alter this PR if it is not aligned properly with the Contribution Guidelines.

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -1504,9 +1504,7 @@
     'name': 'string.regexp.single-quoted.php'
     'patterns': [
       {
-        'comment': 'Support both PHP string and regex escaping'
-        'match': '(?x) \\\\ (?: \\\\ (?: \\\\ [\\\\\']? | [^\'] ) | . )'
-        'name': 'constant.character.escape.php'
+       'include': '#single_quote_regex_escape'
       }
       {
         'captures':
@@ -1524,25 +1522,16 @@
             'name': 'punctuation.definition.character-class.php'
         'end': '\\]'
         'name': 'string.regexp.character-class.php'
-        # 'patterns': [
-        #   {
-        #     # 'include': '#single_quote_regex_escape'
-        #   }
-        # ]
       }
       {
         'match': '[$^+*]'
         'name': 'keyword.operator.regexp.php'
       }
-      # {
-      #   'include': '#single_quote_regex_escape'
-      # }
     ]
-    # 'repository':
-    #   'single_quote_regex_escape':
-    #     'comment': 'Support both PHP string and regex escaping'
-    #     'match': '(?x) \\\\ (?: \\\\ (?: \\\\ [\\\\\']? | [^\'] ) | . )'
-    #     'name': 'constant.character.escape.php'
+  'single_quote_regex_escape':
+    'comment': 'Support both PHP string and regex escaping'
+    'match': '(?x) \\\\ (?: \\\\ (?: \\\\ [\\\\\']? | [^\'] ) | . )'
+    'name': 'constant.character.escape.php'
   'sql-string-double-quoted':
     'begin': '"\\s*(?=(SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER)\\b)'
     'beginCaptures':
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -1504,6 +1504,11 @@
     'name': 'string.regexp.single-quoted.php'
     'patterns': [
       {
+        'comment': 'Support both PHP string and regex escaping'
+        'match': '(?x) \\\\ (?: \\\\ (?: \\\\ [\\\\\']? | [^\'] ) | . )'
+        'name': 'constant.character.escape.php'
+      }
+      {
         'captures':
           '1':
             'name': 'punctuation.definition.arbitrary-repetition.php'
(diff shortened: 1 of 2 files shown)


grammars/php.cson
```diff
@@ -1519,25 +1524,25 @@
             'name': 'punctuation.definition.character-class.php'
         'end': '\\]'
         'name': 'string.regexp.character-class.php'
-        'patterns': [
-          {
-            'include': '#single_quote_regex_escape'
```
50Wliu: So are you saying that this isn't needed, or that it isn't working properly?

bj7: I commented out the redundant sections after I placed the single_quote_regex_escape content into the patterns array as its own item, rather than being a referenced sub-item. I wasn't sure whether I should completely remove the comments or not prior to review and finalization. The code works, just the scoping was off. Does that explain it better?

bj7:
Out of curiosity, what's the status on the review? No rush at all (lots of bigger things to handle), just wondering.

50Wliu:
Currently working through a backlog of some other language PRs and haven't gotten time to tinker around with this one yet.  Don't worry though, I haven't forgotten about it.

bj7:
Ok cool. Seriously, no rush or anything. I was just curious.

50Wliu:
Ok, tried this out locally, and I have a few suggestions:
1. Move `single_quote_regex_escape` to the top of the `patterns` block
2. Un-nest `single_quote_regex_escape` so that it's no longer nested under its own `repository` block and has the same indentation level as `sql-string-double-quoted`.

Making those two adjustments fixed #124 for me, and it looks a bit cleaner than what you did.

In other words...
```diff
:100644 100644 93d4c72... 0000000... M	grammars/php.cson

diff --git a/grammars/php.cson b/grammars/php.cson
index 93d4c72..ef5ba03 100644
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -1504,6 +1504,9 @@
     'name': 'string.regexp.single-quoted.php'
     'patterns': [
       {
+        'include': '#single_quote_regex_escape'
+      }
+      {
         'captures':
           '1':
             'name': 'punctuation.definition.arbitrary-repetition.php'
@@ -1529,15 +1532,11 @@
         'match': '[$^+*]'
         'name': 'keyword.operator.regexp.php'
       }
-      {
-        'include': '#single_quote_regex_escape'
-      }
     ]
-    'repository':
-      'single_quote_regex_escape':
-        'comment': 'Support both PHP string and regex escaping'
-        'match': '(?x) \\\\ (?: \\\\ (?: \\\\ [\\\\\']? | [^\'] ) | . )'
-        'name': 'constant.character.escape.php'
+  'single_quote_regex_escape':
+    'comment': 'Support both PHP string and regex escaping'
+    'match': '(?x) \\\\ (?: \\\\ (?: \\\\ [\\\\\']? | [^\'] ) | . )'
+    'name': 'constant.character.escape.php'
   'sql-string-double-quoted':
     'begin': '"\\s*(?=(SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER)\\b)'
     'beginCaptures':
```

bj7:
Ok then I will do that this weekend!

50Wliu:
Can you please add a spec for this to prevent this behavior from regressing in the future?  Thanks!

bj7:
Oh yes! Sorry, totally skipped my mind.

bj7:
So I'm having a bit of an issue getting the spec to work. I attempted to use the most simple case that causes a failure prior to the fix, but that works after the fix:
```php
<?php
'/\[/'
```
This is an image with the fix applied, and the html parsing shown. It's very simple and obvious what the parsed classes should be.
image
I wrote up the spec based on this working case:
```coffeescript
it 'should tokenize single quoted string regex with escaped bracket', ->
        tokens = grammar.tokenizeLines "<?php\n'/\[/'"

        expect(tokens[1][0]).toEqual value: '\'/', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.regexp.single-quoted.php', 'punctuation.definition.string.begin.php']
        expect(tokens[1][1]).toEqual value: '\[', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.regexp.single-quoted.php', 'constant.character.escape.php']
        expect(tokens[1][2]).toEqual value: '/\'', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.regexp.single-quoted.php', 'punctuation.definition.string.end.php']
```
However, running the spec always fails because the tokenizer determines that `\[` is a punctuation class, even though in a real PHP file it parses correctly as a constant escape class:
image
I have no clue as to why the PHP parser works, but the spec parser fails to tokenize correctly. My only assumption at this point is that the spec tokenizer is somehow different from the one used when parsing PHP files, and is demonstrably broken. If you could give me any insight on how I might continue working on this, I would greatly appreciate it.

bj7:
Oh wow I did not think of that, thank you so much. It now works as expected.

50Wliu:
Thanks!

bj7:
Thank you for your help!

PR #141, opened 2016-06-15
indent all lines between parentheses

Hikonobu:
Refs #63 

1. A `(` did not trigger an indent except for `array(`
2. A `) {` did not trigger an outdent

### DIFF
--- a/settings/language-php.cson
+++ b/settings/language-php.cson
@@ -3340,8 +3340,8 @@
     'decreaseIndentPattern': '(?x)\n\t    ^ (.* \\*/)? \\s*\n\t    (\\)+)\n\t'
 '.source.php:not(.string)':
   'editor':
-    'increaseIndentPattern': '(?x)\n\t    (   \\{ (?! .+ \\} ) .*\n\t    |   array\\(\n\t    |   (\\[)\n\t    |   ((else)?if|else|for(each)?|while|switch) .* :\n\t    )   \\s* (/[/*] .*)? $'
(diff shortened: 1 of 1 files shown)


50Wliu:
While you're at it, can you also please tidy these regexes up?  See https://github.com/atom/language-sass/blob/50e674c250682ce68b4200e50d7ccb85c82fe241/settings/language-sass.cson#L4-L14 for an example.

Hikonobu:
I will.
Should I tidy all regexes up in the file?

50Wliu:
Yes, that would work.

charlesqwu:
Thanks for the fix!  While your fix has not been merged, I just copied and pasted the following lines into my atom's config,cson file -- it works fine. This is just a tip for other people who're having the same problem.
<code>
'.source.php:not(.string)':
   'editor':
     'increaseIndentPattern': '(?x)\n\t    (   \\{ (?! .+ \\} ) .*\n\t    |   \\(\n\t    |   (\\[)\n\t    |   ((else)?if|else|for(each)?|while|switch) .* :\n\t    )   \\s* (/[/*] .*)? $'
     'decreaseIndentPattern': '(?x)\n\t    ^ (.* \\*/)? \\s*\n\t    (\n\t        (\\})         |\n\t        (\\)+([;,]|\\s*\\{))    |\n\t        (\\]\\)*([;,]|$))     |\n\t        (else:)      |\n\t        ((end(if|for(each)?|while|switch));)\n\t    )\n\t'
<code>

50Wliu:
Hi @charlesqwu, could you test this out a bit more and see if there's any problems with it?  That's all that this PR is waiting on.

charlesqwu:
Issue #63 has two parts: 1. auto-indent issue related to () and [] -- the commit #141 fixes this part very well. I have tested many conditions related to this part, and it works fine. 2. auto-indent issue related to heredoc and newdoc. The commit #141 does not fix this second part.

However, most PHP developers would experience the second issue a lot less frequently than the first issue, and so I would suggest that we merge the commit #141 as soon as possible.

PR #146, opened 2016-07-15
Change "var_dump" to be reachable only in a php tag

excellent6:
Change "var_dump" to be reachable only in a php tag. It can be used in the whole file.


50Wliu:
Thanks!

Issue #156, opened 2016-09-25
try...catch snippet namespace


daniel-centore:
The current try...catch snippet looks like this:

```
try {

} catch (Exception $e) {

}
```

This code does not work if the file is in a non-default namespace. The default snippet should be written to work in any namespace:

```
try {

} catch (\Exception $e) {

}
```

50Wliu:
Fixed by #157.

PR #157, opened 2016-09-26
Fixed: Block try...catch now uses \Exception instead of Exception

eliasib13:
This solves #156 issue :)

### DIFF
--- a/snippets/language-php.cson
+++ b/snippets/language-php.cson
@@ -109,7 +109,7 @@
     'body': 'throw new $1Exception(${2:"${3:Error Processing Request}"}${4:, ${5:1}});\n$0'
   'try …':
     'prefix': 'try'
-    'body': 'try {\n\t$0\n} catch (${1:Exception $e}) {\n\t$2\n}\n'
+    'body': 'try {\n\t$0\n} catch (${1:\\\\Exception $e}) {\n\t$2\n}\n'
   'while …':
     'prefix': 'while'
     'body': 'while (${1:$a <= 10}) {\n\t${0:# code...}\n}'
--- a/snippets/language-php.cson
+++ b/snippets/language-php.cson
@@ -106,7 +106,7 @@
     'body': 'echo $this->$0'
   'Throw Exception':
     'prefix': 'throw'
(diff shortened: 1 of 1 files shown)


Ingramz:
For consistency, `throw` snippet should get the same change.

However in general I have mixed feelings about all of this, but I am not going to say it's by any means bad.

50Wliu:
> However in general I have mixed feelings about all of this

Can you elaborate?

Ingramz:
I just think it's rather obvious that when using namespaces, you are somewhat supposed to keep track of this yourself. Adding backslash in front to denote the root isn't bad on it's own but catching a generic Throwable/Exception/Error most of the time is. For quick and dirty development I suppose it is ok. The snippets should try to encourage changing the name of Exception class though.

50Wliu:
> The snippets should try to encourage changing the name of Exception class though.

So are you saying it should be `\\\\$1Exception` instead?

Ingramz:
No, not really. My concern is/was that by making the snippet work, it's not giving user enough incentive to use a better name for the exception that is going to be thrown/caught. I've now concluded that this is something we can't protect against.

The way how certain parts of the snippet are selected is more related to usability of the snippets, it would be better to ask others on this subject.

For instance I would prefer myself if the whole exception name + backslash was selected as a whole and then if I skipped that with tab, it would only highlight the exception name, leaving backslash intact. How the variable name should be selected is hard to tell, maybe like this:
**\Exception $e**
\\**Exception $e**
\\**Exception** $e
\Exception **$e**

This should allow overwriting as soon as you have reached the desired level, but this is not the only solution.

joehoyle:
Im not sure this is a good default, in a file that doesn't have  namespace, the `\` in unnecessary - and even in a file that _does_ have a namespace, I'd say typically it's best private to `use` the `Exception`  class at the top of the file. 

Personally, I don't think the snippet needs to work in every context, just as I can't use a try catch snippet in plenty of other contexts, it's not the job of the snippets to be contextual like that.

AlexNodex:
@joehoyle While the `\` is unnecessary in files that don't have namespaces, it does no harm whatsoever. It visually promotes to the programmer that it's a root `Exception`. Namespaced files are much more prevelant than non namespaced files and in my experience it's good to know what's in the root namespace and what isn't.

PR #158, opened 2016-09-26
Add completions for PHP functions with descriptions, return types and and params

joehoyle:
This significantly improves the UX for PHP projects by adding
autocompletion configs for PHP function return values, parameter names,
descriptions and "since version" to the description.

The list of functions is generated from the JetBrains PHP-Stubs repo
(https://github.com/JetBrains/phpstorm-stubs)

This looks like this:


joehoyle:
I also added PHP constants and class names.

Ingramz:
Good job @joehoyle, but can we somehow automate this in a way that anyone can update these programmatically in the future?

Similar to https://github.com/atom/language-less/blob/master/update.coffee

joehoyle:
@Ingramz sure, I have a script to do this in PHP which builds the whole `settings.json` file - I can add that too - it's in PHP so anyone wanting to update it will need that of course, this is because the PHP Parser is written in PHP.

joehoyle:
@Ingramz ok I added the `update` project this it's own readme too.

joehoyle:
@Ingramz any update here?

Ingramz:
@joehoyle very cool! Only pitfall that I am seeing here is that JSON file is re-generated and not simply modified. By all means this is better than no script.

Also I have no clue if using JSON instead of CSON is desired, but I understand that's because of PHP limitations. I also thought of rewriting the scripts to pure javascript, which seems to be possible (perhaps using [this](https://github.com/glayzzle/php-parser) and [this](https://github.com/Andifeind/docblock)). This is not required, just something that might be easier to maintain for javascript-oriented community.

In addition to that, some languages have an `autocomplete-<some language>` package instead of the grammar and autocomplete support being in the same repo. I don't know what the policy on that is though.

50Wliu:
> the reason for that is that people might try to modify the JSON directly, which can be easily overwritten by the script.

I'm for all manual changes getting wiped out so that the script is deterministic.

Ingramz:
I still think it would be better to solve it like this https://github.com/atom/language-less/blob/master/update.coffee#L31-L35

joehoyle:
Ok, happy to do that!

joehoyle:
However, re: 

> In addition to that, some languages have an autocomplete-<some language> package instead of the grammar and autocomplete support being in the same repo. I don't know what the policy on that is though.

I feel like this is a somewhat important point and probably shouldn't continue this effort until this is decided.

50Wliu:
I'm not too sure whether this should belong here or if we're trying to move all autocompletions into dedicated autocomplete packages.  /cc @joefitzgerald

PabloJoan:
IT BELONGS HERE!!

These are core php functions. not support of a library or a framework.

Issue #155, opened 2016-09-22
PHP EOP, xml content


MaksOuw:
In PHP, when using <<<EOP var type with XML content, the syntax colorization is not right. See screenshot, line 237 (in the screenshot).

It's because of the XML first line declaration, which is endded with "?>", which is the PHP closing tag too. When I remove it, the syntax color is OK.

As you can see too, if there's underscore char in XML tag, the whole word after the underscore, with the closing tag, is not colorized (see <masque_saisie> and <masque_rendu> on the screen, line 239-240)

![capture du 2016-09-22 15-47-35](https://cloud.githubusercontent.com/assets/4067614/18750581/06d4dcc6-80dc-11e6-99d1-02af4e024541.png)

Ingramz:
@50Wliu this is a duplicate of other heredoc-not-working issues.

50Wliu:
Looks like this is a duplicate of #57 then.

Issue #133, opened 2016-04-25
Heredoc issue


Asdra:
Hi, I noticed another issue with heredoc syntax, it seems that highlighting breaks when it meets a `'`, here is an example :

![bug_atom_1](https://cloud.githubusercontent.com/assets/7705276/14778787/3cd848bc-0ad5-11e6-8bbf-39f23aef244d.png)

The entire file is not highlighted after this bug.

Ingramz:
This is not a completely different issue, just a side effect of the current issue at hand. Everything after the first `<` is parsed like source code, orange being constants. If heredoc worked, we wouldn't have this issue.

At the pointed location you begin a string which will be never closed, so it interprets rest of the file as string. To work around this, you could end the heredoc like this: `HEREDOC; // '` which will close the opened string.

Aldlevine:
:+1:

sscotti:
Anyone ever make any progress on this.  I just discovered and it is pretty annoying.  If I want to include CSS in the HEREDOC string the "#" for Id's breaks the formatting, comments out everything that follows.

50Wliu:
Fixed by #184.

Issue #113, opened 2015-12-19
Heredoc is not processed.


IngwiePhoenix:
``` php
<?php
$str = <<<EOT
This is a foo.
    He likes bar.
        But bar likes baz.
EOT;
```

This is not recognized, but yet a simplified variant from the official PHP docs. Proper heredoc support is required since larger string sections in PHP are best done using that. For instance, I use it to store translated paragraphs.

rugk:
There seem to be several [issues with Heredoc](https://github.com/atom/language-php/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+Heredoc+). Possibly this is a duplicate of https://github.com/atom/language-php/issues/10.

IngwiePhoenix:
Yeah, looks like I am not the only one being unable to heredoc their stuff (:

Is a possible fix in sight?

sbrl:
I don't think that this is a duplicate of #10, because heredoc isn't processed correctly even when it doesn't contain a closing PHP tag.

50Wliu:
Fixed by #184.

Issue #78, opened 2015-05-26
Bug folding PHP heredocs-style strings.


DionisiusMayr:
Bug: create a function that has a heredoc string on it and a couple more lines after the end of the heredoc, then try to fold the function. It will stop at the word you used as the heredoc delimiter -since it need to be placed at the start of a line.

I'm in a hurry here so I can't upload images, but there is a better description of the bug in this post that I made yesterday:
https://discuss.atom.io/t/possible-bug-folding-php-heredocs-style-strings/17154

50Wliu:
Thanks for taking the time to contribute!

We noticed that this is a duplicate of https://github.com/atom/atom/issues/3442. You may want to subscribe there for updates.

Because we treat our issues list as [the Atom team's backlog](https://en.wikipedia.org/wiki/Scrum_(software_development)#Product_backlog), we close duplicates to focus our work and not have to touch the same chunk of code for the same reason multiple times. This is also why we may mark something as duplicate that isn't an exact duplicate but is closely related.

For information on how to use GitHub's search feature to find out if something is a duplicate before filing, see [the How Can I Contribute? section of the Atom CONTRIBUTING guide](https://github.com/atom/atom/blob/master/CONTRIBUTING.md#how-can-i-contribute).

PR #184, opened 2017-02-19
temporary fix for heredocs

UziTech:
### Description of the Change

Temporary fix for heredocs explained in [#77 (comment)](https://github.com/atom/language-php/pull/77#issuecomment-140815381)

### Alternate Designs

#77 has been waiting on a fix for over a year on another package that doesn't look like it is being worked on. Fixing the underlying problem would be the optimal solution but this will provide a fix in the mean time.

### Benefits

Heredocs will work as expected including sub-styling with `<<<HTML`, `<<<SQL`, etc.

### Possible Drawbacks

None that I can see.

### Applicable Issues

#10, #57, #63, #78, #113, #133, #155


dustinwilson:
Thanks for doing this pull request. I tested this, and interpolation isn't working.

Because of the changes you made the interpolation stuff at [line 475](https://github.com/UziTech/language-php/blob/cc576387c51e89197e964e4aabfb89be4eeb6a32/grammars/php.cson#L475) doesn't do anything.

However, if you put the interpolation stuff like so after [line 713](https://github.com/UziTech/language-php/blob/cc576387c51e89197e964e4aabfb89be4eeb6a32/grammars/php.cson#L713) it should work (and probably for the other ones if you repeat it):

```CoffeeScript
{
  'begin': '(<<<)\\s*([\'"]?)([a-zA-Z_]+[a-zA-Z0-9_]*)(\\2)'
  'beginCaptures':
    '1':
      'name': 'punctuation.definition.string.php'
    '3':
      'name': 'keyword.operator.heredoc.php'
  'end': '^(\\3)\\b'
  'endCaptures':
    '1':
      'name': 'keyword.operator.heredoc.php'
  'patterns': [
    {
      'include': '#interpolation'
    }
  ]
}
```

UziTech:
Thank you @dustinwilson. I noticed that last night but I wasn't sure how to fix it.

UziTech:
The only other part I'm not 100% sure on is in the HTML heredoc test it looks like `<strong>rainbows</strong>` just gets the `text.html` scope as if it is a string.

In the other pull request (#77) it looks like the tag should have additional scopes

I'm not sure what to do (if anything) about that.

Everything seems to actually work in Atom

UziTech:
It looks like the reason it is not adding additional scopes to the tag is because the 'language-html' package is not activated in the tests.

That should be fine as long as it gets the `text.html` scope the 'language-html' package should take over in atom so there should be no need to test that in this package

50Wliu:
In that test, add a waitsForPromise like follows:
```coffee
waitsForPromise ->
  atom.packages.activatePackage('language-html')
```

The rest of the test will need to be wrapped in a `runs` function, but then it should work.

UziTech:
@50Wliu Do you think we need to add the language-html scopes to the test?

I could see an issue where language-html changes something that breaks these tests and it would be hard to figure out why they broke.

I do test for the `text.html` scope on anything that is supposed to be html. Adding the `text.html` scope should be enough for this package.

50Wliu:
My worry is that what if someone removes the HTML include later on?  Then we have no way of knowing that it broke just from the specs.

I wholeheartedly agree that the whole cross-language-specs thing is very fragile though - if you have a potential solution, I would love to hear it!

UziTech:
True. I don't think we are going to get around this one.

Do you want me to add tests for the other languages as well?

50Wliu:
Yes please.  Just a simple test will suffice.

UziTech:
lol apparently this is happening for language-css.

The tests for css can't pass because the stable channel has a different scope for curly braces than the beta channel.

50Wliu:
Create the tests so that they pass using the latest version of the language package (looks like that's what you did).

UziTech:
you want them to pass the latest stable version? Then they will continue to be broken.

If I change them to pass the beta version then they will at least pass when the beta version becomes stable

50Wliu:
Yes, what you did is fine.  When I say latest, I mean "the latest version available", not "the version available on stable".

grammars/php.cson
```diff
@@ -470,15 +470,8 @@
   'heredoc':
     'patterns': [
       {
-        'begin': '(?=<<<\\s*("?)([a-zA-Z_]+[a-zA-Z0-9_]*)(\\1)\\s*$)'
+        'begin': '(?=<<<\\s*("?)([a-zA-Z_]+[a-zA-Z0-9_]*)(\\1)[ \\t]*$)'
```
50Wliu: Why was this changed?

grammars/php.cson
```diff
@@ -487,233 +480,506 @@
...
           }
         ]
       }
     ]
-    'repository':
-      'heredoc_interior':
+  'heredoc_interior':
+    'patterns': [
+      {
+        'begin': '(<<<)\\s*("?)(HTML)(\\2)([ \\t]+)?$'
```
50Wliu: The `[ \\t]` should be changed to use `\\s` instead.

spec/php-spec.coffee
```diff
@@ -488,6 +488,690 @@ describe 'PHP grammar', ->
...
+        expect(tokens[9][0]).toEqual value: '    console.log(x*x);', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.unquoted.heredoc.nowdoc.php']
+        expect(tokens[10][0]).toEqual value: '  });', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.unquoted.heredoc.nowdoc.php']
+        expect(tokens[11][0]).toEqual value: '}', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.unquoted.heredoc.nowdoc.php']
+        expect(tokens[12][0]).toEqual value: 'GITHUB', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'string.unquoted.heredoc.nowdoc.php', 'keyword.operator.heredoc.php']
+        expect(tokens[12][1]).toEqual value: ';', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'punctuation.terminator.expression.php']
+
+      it 'should tokenize a heredoc with embedded HTML and interpolation correctly', ->
+        waitsForPromise ->
+          atom.packages.activatePackage('language-html')
+        runs ->
```
50Wliu: Add a newline before the runs

grammars/php.cson
```diff
@@ -487,233 +480,506 @@
         ]
       }
       {
-        'begin': '(?=<<<\\s*(\'?)([a-zA-Z_]+[a-zA-Z0-9_]*)(\\1)\\s*$)'
+        'begin': '(?=<<<\\s*\'([a-zA-Z_]+[a-zA-Z0-9_]*)\'[ \\t]*$)'
         'end': '(?!\\G)'
         'name': 'string.unquoted.heredoc.nowdoc.php'
```
50Wliu: Shouldn't this just be `string.unquoted.nowdoc.php` instead?

grammars/php.cson
UziTech: according to the [original pull request](https://github.com/atom/language-php/pull/77/commits/eb4ae2656bd8a427d2026f082ba2c62da3d62654#diff-9972c3dda8fb0e2f637ca073c25e7e6dR439)
```
[ \\t] was used instead of \\s because \\s will match a newline
```

grammars/php.cson
UziTech: that's what I was thinking but I didn't know if there was a reason for the `.heredoc` (styling or whatever)

I can change it if you want

grammars/php.cson
UziTech: see https://github.com/atom/language-php/pull/184#discussion_r105465668

UziTech:
Seriously this is a never ending battle. 

Now `language-json` scopes changed in beta. 

I'm thinking testing the other packages' scopes is not going to be worth the hassle of having to update the tests any time one of the other language packages updates.

grammars/php.cson
50Wliu: I highly doubt that as the begin pattern can only match a single line.  If you change it, do the specs still pass?

UziTech: ya everything looks fine when it is `\\s`

50Wliu:
> I'm thinking testing the other packages' scopes is not going to be worth the hassle of having to update the tests any time one of the other language packages updates.

Maybe we could test just one of the scopes (probably meta ones) that isn't likely to change using `.toContain`.

UziTech:
I changed the tests to check for the token value and any scopes that were added by language-php. That way if the included languages are removed then the token values won't match but it won't fail if other packages change their scopes only if they tokenize the string differently, which is unlikely.

grammars/php.cson
```diff
@@ -487,233 +480,506 @@
...
+            'name': 'keyword.operator.heredoc.php'
+          '3':
+            'name': 'invalid.illegal.trailing-whitespace.php'
+        'contentName': 'text.html'
+        'end': '^(\\2)\\b'
+        'endCaptures':
+          '0':
+            'name': 'punctuation.section.embedded.end.php'
+          '1':
+            'name': 'keyword.operator.heredoc.php'
```
UziTech: do you want me to change this to `keyword.operator.nowdoc.php`

50Wliu: (assuming you meant for `nowdoc_interior` and not `heredoc_interior`) yes please.

grammars/php.cson
```diff
@@ -487,233 +480,506 @@
...
           }
         ]
       }
     ]
-    'repository':
-      'heredoc_interior':
+  'heredoc_interior':
+    'patterns': [
+      {
+        'begin': '(<<<)\\s*("?)(HTML)(\\2)(\\s+)?$'
```
50Wliu: One last thing.  I believe that the `(\\s+)?` can be changed to simply `(\\s*)` - can you confirm?

50Wliu:
Thanks!

50Wliu:
Current ETA is Atom 1.17.0.  You can manually install language-php using `apm install language-php` though that is **highly** discouraged (and will display deprecation warnings starting with Atom 1.16.0).

timbotron:
@50Wliu Thank you for the info! I think I'll just wait. :)

AndreasWebdev:
It seems like this was not fixed. I am on 1.17.2

image

jens1o:
The closing character must not be indented, so make sure `FDF;` is at the very first column.

PR #185, opened 2017-02-21
Closure snippet

arandilopez:
### Description of the Change
Snippet for closure

### Benefits

By typing `f` an then press `tab` you get:

```php
<?php

function ($var = '') {
    // code...
}
```

> Same as language-javascript set for closures.

### Possible Drawbacks

NOPE

### Applicable Issues

NOPE

### DIFF
--- a/snippets/language-php.cson
+++ b/snippets/language-php.cson
@@ -56,7 +56,7 @@
   'function …':
     'prefix': 'fun'
     'body': '${1:public }function ${2:FunctionName}(${3:$${4:value}${5:=\'\'}})\n{\n\t${0:# code...}\n}'
-  'annonymous function':
+  'anonymous function':
     'prefix': 'f',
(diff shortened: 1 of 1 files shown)


Ingramz:
One-letter snippet feels to me a bit too short, but I understand that `fun` has been already taken. One possibility would be to use `fn` but then it might be too close to `fun`. I'd like to hear what others think.

snippets/language-php.cson
```diff
@@ -56,6 +56,9 @@
   'function …':
     'prefix': 'fun'
     'body': '${1:public }function ${2:FunctionName}(${3:$${4:value}${5:=\'\'}})\n{\n\t${0:# code...}\n}'
+  'annonymous function':
```
50Wliu: `anonymous`

arandilopez: oh man, thanks

Issue #135, opened 2016-05-07
Highlight single-line docblocks


daltones:
I think it's common write `@var` tags in a single-line docblock for class properties. PHP Documentor actually suggests an example here: https://www.phpdoc.org/docs/latest/references/phpdoc/tags/var.html.

I don't know why TextMate based syntaxes have this behavior, but maybe atom should accept single-line docblocks at all and highlight all `@` tags.

``` php
/**
 * @var type Description
 */
protected $property;
```

``` php
/** @var type Description */        // No highlight here!
protected $property;
```

kafoso:
Agreed. Would very much appreciate this.

May also be useful for adding references to ordinary variables in single-line comments, e.g. in loops as this one:

```php
foreach ($users as $user) { // @var $user My\Application\Entity\User
    // ...
}
```

`@override` also comes to mind, which isn't highlighted, either.

I'm aware I'm using a single-line comment in the above example. However, I am very strongly against using multiline comments when a single line will suffice. Simply because of debugging annoyances. Below code will break down when wrapped in a multi-line comment while the `@var` reference is wrapped in a multi-line comment, too:

```php
/*
foreach ($users as $user) { /** @var $user My\Application\Entity\User */
    // ...
}
*/
```

Of course, restrictions should be enforced so that single-line comments won't get sporadially colored. E.g.:

1. Start with two slashes (not inside a string).
2. One space character (and not tab).
3. Notation, e.g. `@var` or `@override`.
4. One space character (and not tab).
5. The remainder of the comment, formatted as usual.

50Wliu:
Fixed by #191.

PR #191, opened 2017-04-19
Inline phpdoc

UziTech:
### Description of the Change

highlight inline phpdoc correctly

```php
/** @var string Variable */
```

### Alternate Designs

We might want to also support single line comments https://github.com/atom/language-php/issues/135#issuecomment-270901684

### Benefits

highlight inline phpdoc correctly

### Possible Drawbacks

none

### Applicable Issues

#135

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -178,12 +178,15 @@
   'comments':
     'patterns': [
       {
-        'begin': '/\\*\\*(?:(?:(?:#@\\+)?\\s*$)|\\s+)'
-        'captures':
+        'begin': '(/\\*\\*)(((#@\\+)?\\s*$)|\\s+)'
+        'beginCaptures':
+          '1':
+            'name': 'punctuation.definition.comment.php'
+        'comment': 'This now only highlights a docblock if the first line contains only /**\n\t\t\t\t\t\t\t\t- this is to stop highlighting everything as invalid when people do comment banners with /******** ...\n\t\t\t\t\t\t\t\t- Now matches /**#@+ too - used for docblock templates: https://codingexplained.com/coding/php/how-to-use-docblock-templates-in-phpdoc'
+        'end': '\\*/'
+        'endCaptures':
           '0':
             'name': 'punctuation.definition.comment.php'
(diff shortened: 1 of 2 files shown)


grammars/php.cson
```diff
@@ -192,6 +192,19 @@
         ]
       }
       {
+        'begin': '/\\*\\*\\s+'
+        'captures':
```
50Wliu: Use beginCaptures and endCaptures

grammars/php.cson
```diff
@@ -192,6 +192,19 @@
         ]
       }
       {
+        'begin': '/\\*\\*\\s+'
```
50Wliu: Don't capture the trailing/leading whitespace

grammars/php.cson
```diff
@@ -178,12 +178,15 @@
   'comments':
     'patterns': [
       {
-        'begin': '/\\*\\*(?:#@\\+)?\\s*$'
-        'captures':
-          '0':
+        'begin': '(/\\*\\*)(((#@\\+)?\\s*$)|\\s+)'
```
UziTech: I'm not sure if we should keep the #@+ syntax. 

The site that was referenced in the comment no longer exists and I could only find one site that mentioned the syntax and referenced the same broken link that was in the comment

50Wliu: If there's no documentation on it then I think it's safe to remove.

grammars/php.cson
```diff
@@ -178,12 +178,16 @@
   'comments':
     'patterns': [
       {
-        'begin': '/\\*\\*(?:#@\\+)?\\s*$'
-        'captures':
-          '0':
+        # This only highlights a docblock if the first line contains only two *s
+        # to stop highlighting everything as invalid when people do comment banners with /******** ...
+        'begin': '(/\\*\\*)(?=$|\\s+)'
```
50Wliu: Is the `$` necessary?

UziTech: I guess not the \s should match a \n at the end.

We just don't want to match /***

PR #208, opened 2017-06-09
Tokenize parentheses

### LINKED ISSUE #47: Highlighting parentheses
I tried taking a stab at adding this in the `php.cson` file - but didn't get far.

Things like `if(true)` or `case($foo)`, or native functions `in_array()` - the parentheses are not wrapped in a span with a class(es), meaning we can't target them in syntax themes.

Example from the dom on an if statement checking a `file_exists`: [source](https://www.dropbox.com/s/uilh6kiy3ip7eul/Screenshot%202014-12-03%2017.07.06.png?dl=0)

Can this be added?  I'll keep digging through the language file in the meantime.

### LINKED ISSUE #109: Method calls in strings (string parsing) are incorrectly highlighted.
Hello

Currently variables inside double quoted strings are correctly highlighted:

``` php
$string = "
    Do {$this->test} this.
";
```

However, the syntax highlighting becomes somewhat awkward when turning it into a method call:

``` php
$string = "
    Do {$this->test()} this.
";
```

`test()` is then suddenly highlighted as the rest of the string instead of in a different color.

Thanks in advance

### LINKED ISSUE #202: catch highlighting doesn't work on multiple lines
### Description

Colorizing of PHP catch clause seems broken if the trailing close-parenthesis is on a different line:

On a different line (fails):
image

On the same line (works fine):
image

### Steps to Reproduce

1. Create a new file. Save it with the name "anything.php". Type in the following text:

```php
<?php

function f() {
  try {} catch (Exception $e
  ) {
      $x = 17;
    }
}

function g() {
}
```

**Expected behavior:** It should colorize the word "Exception", and also the word "function" in the declaration of g.

**Actual behavior:** Everything after the catch has completely lost all of its colorization.

**Reproduces how often:** 100%

### Versions

```
$ atom --version
Atom    : 1.16.0
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0

$ apm --version
apm  1.16.1
npm  3.10.5
node 6.9.5 x64
python 2.7.10
git 2.13.0
```

OS: macOS Seirra 10.12.4

### Additional Information

Incidentally, I observe the same issue in VSCode.

50Wliu:
### Requirements

### Description of the Change

Adds `punctuation.definition.*.begin/end.bracket.round.php` scopes to all parentheses.  While doing this I uncovered a few bugs and fixed those along the way.  Notably, function and method calls now receive the `entity.name.function.php` scope, catch statements receive proper classes when there is no space between the `catch` and the opening parentheses, and the scope resolution operator now works as expected, rather than always being tokenized as a goto label.  Specs have been added for most/all of the code I touched.

### Alternate Designs

None.

### Benefits

If your syntax theme highlights parentheses, all parentheses should be properly highlighted, and scopes won't be cut off abruptly.

### Possible Drawbacks

I rewrote/reorganized some regexes.  Perhaps those will have some regressions.

### Applicable Issues

Fixes #47
Fixes #109
Fixes #202

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -531,12 +531,12 @@
           '6':
             'name': 'support.function.construct.php'
           '7':
-            'name': 'punctuation.definition.array.begin.php'
+            'name': 'punctuation.definition.array.begin.bracket.round.php'
         'contentName': 'meta.array.php'
         'end': '\\)'
         'endCaptures':
           '0':
-            'name': 'punctuation.definition.array.end.php'
+            'name': 'punctuation.definition.array.end.bracket.round.php'
         'name': 'meta.function.argument.array.php'
         'patterns': [
           {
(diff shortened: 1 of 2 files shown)


Ingramz:
Good work, @50Wliu !

I'll try it out on Blade to see if bracket matching can be solved with these changes.

50Wliu:
@Ingramz I'm merging this now so I can get this out in beta and work on other issues, but definitely file issues if you find bugs/areas for improvement.

PR #214, opened 2017-06-10
Support 'use' in traits

### LINKED ISSUE #127: Incorrect highlighting for trait uses inside classes.
Hello

It seems a couple of things are going wrong when dealing with use statements that contain aliases:

``` php
class Foo
{
    use TraitOneMethod {
        test as private test1;
        TraitOneMethod::method insteadof TraitOneMethod;
    }
}
```

Here, `test`, `as`, `private` and `test1` are all marked as separate `.other.namespace.use` items. `TraitOneMethod` is correctly determined to be a class name, but `method`, `insteadof` and `TraitOneMethod` are all seen as separate `constant`s.

Thanks in advance

### LINKED ISSUE #200: Support trait overwrites
image

Code: 
```php
<?php
trait A {
    public function smallTalk() {
        echo 'a';
    }
    public function bigTalk() {
        echo 'A';
    }
}

trait B {
    public function smallTalk() {
        echo 'b';
    }
    public function bigTalk() {
        echo 'B';
    }
}

class Talker {
    use A, B {
        B::smallTalk insteadof A;
        A::bigTalk insteadof B;
    }
}

class Aliased_Talker {
    use A, B {
        B::smallTalk insteadof A;
        A::bigTalk insteadof B;
        B::bigTalk as talk;
    }
}

50Wliu:
### Requirements

### Description of the Change

Adds support for the `use` directive in classes.  In order to differentiate this from the existing `use` namespace syntax, I had to extend the class begin/end to cover the entire class, similar to what Java does.

/cc @jens1o, @Ingramz, @UziTech - any comments on my scope naming?  Anything that I missed?  Mostly I just made sure that the examples at http://php.net/manual/en/language.oop5.traits.php all worked.

### Alternate Designs

None.

### Benefits

First off, the obvious benefit: trait use syntax will be tokenized correctly.  Next, the not-so-obvious benefit: the class regex will be setting a precedent for how to cover entire blocks, rather than just the section up to the block start.

### Possible Drawbacks

I don't think there will be any.

### Applicable Issues

Fixes #127
Fixes #200

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -212,7 +212,7 @@
                 '''
                 'captures':
                   '1':
-                    'name': 'keyword.operator.use-as.php'
+                    'name': 'keyword.other.use-as.php'
                   '2':
                     'patterns': [
                       {
@@ -231,7 +231,7 @@
                 'match': '(?i)\\b(insteadof)\\s+([a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*)'
                 'captures':
                   '1':
-                    'name': 'keyword.operator.use-insteadof.php'
+                    'name': 'keyword.other.use-insteadof.php'
                   '2':
                     'name': 'support.class.php'
               }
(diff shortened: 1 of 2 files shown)


50Wliu:
Also, here's what it looks like now:
trait-use

jens1o:
I think that's all. Thanks for your effort! :)

UziTech:
Lgtm

PR #216, opened 2017-06-11
Tokenize the special 'class' keyword when used in scope resolution

### LINKED ISSUE #125: Possibly incorrect highlighting on PHP 5.5 'class' keyword for classes.
Hello

The new class keyword introduced in PHP 5.5 is currently being highlighted as 'other constant', though it is really more of a keyword [1]. For example:

``` php
$a = self::class;
$b = static::class;
$c = parent::class;
$d = Foo::class;
```

Thanks in advance

[1] https://wiki.php.net/rfc/class_name_scalars

50Wliu:
### Requirements

### Description of the Change

Does exactly what the title says.

### Alternate Designs

`constant.language.class.php` was considered, but constants generally don't change on context, whereas `class` does.

### Benefits

Scope will be more correct.

### Possible Drawbacks

None.

### Applicable Issues

Fixes #125

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -2317,6 +2317,14 @@
         ]
       }
       {
+        'match': '(?i)(::)\\s*(class)\\b'
+        'captures':
+          '1':
+            'name': 'keyword.operator.class.php'
+          '2':
+            'name': 'keyword.other.class.php'
+      }
+      {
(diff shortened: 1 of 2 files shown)


50Wliu:
Think this will be my last PR for language-php this time around.

PR #217, opened 2017-06-12
Split PHP grammar into separate HTML and PHP portions

### LINKED ISSUE #182: Don't fallback to  'text.html.basic' if starting tag is not found
The reason is that packages that work with parts of a php file's code (ex. code peek) do not get syntax highlighting. 

A possible solutions for packages is to add `<?` and remove it in case the package writes the content back to the file.

Any other ideas?

50Wliu:
### Requirements

### Description of the Change

This PR splits the PHP grammar into two distinct files: `html.cson`, responsible for detecting PHP tags, and `php.cson`, responsible for the actual PHP grammar.  This separation allows other packages to include just the PHP grammar, something that wasn't possible beforehand.

### Alternate Designs

No alternatives were considered.

### Benefits

Packages such as language-gfm and code-peek will be able to tokenize PHP even if the code is not wrapped with `<?php ?>` tags.  First line matches will actually be useful.

### Possible Drawbacks

I don't foresee any.

### Applicable Issues

Fixes #182


50Wliu:
I attempted to merge with master in a1fbacf, however given the scope of the changes many of them were applied incorrectly (such as comma matching being added to `interpolation` instead of the top-level `patterns`...).  Hopefully I've sorted out all the inconsistencies.

Issue #247, opened 2017-07-28
Give `clone` a different scope.


jens1o:
### Prerequisites

* [X] I'm using vscode.

### Description

`clone` has the scope `storage.type.php`, while it isn't really a type, but a keyword, and thus I want to style it differently.

Maybe I can start contributing here?

50Wliu:
> Maybe I can start contributing here?

Sure!

50Wliu:
Closed via #248.

PR #248, opened 2017-07-28
move clone operator out of storage.modifier

jens1o:
### Requirements

### Description of the Change
see #247 

The `clone` operator is not a storage type, but instead a keyword, so I moved it out. 

### Alternate Designs
Move it to a different location?

### Benefits
More precise syntax highlighting.

### Possible Drawbacks
Theme makers may have to restyle it, but most themes just use the `keyword` global scope, so most of them are not affected.

### Applicable Issues

#247

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -1947,7 +1947,7 @@
       }
       {
         'match': '(?i)\\bclone\\b'
-        'name': 'keyword.operator.clone.php'
+        'name': 'keyword.other.clone.php'
       }
       {
         'match': '\\.=?'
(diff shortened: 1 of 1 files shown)


jens1o:
Noob question: What does `\\b` do? Do I need it at the end of the expression?

50Wliu:
`\b` means "make sure there's a word boundary".  In other words, make sure that either there is a word character followed by a non-word character, or a non-word character followed by a word character.

Yes, you need it (or a similar expression) to prevent something like `cloned` to match.

jens1o:
like this?

50Wliu:
I think this is better as `keyword.other.clone.php`.  `keyword.operator` is usually for character operators (`||`, `=`).

jens1o:
Okay.

PR #249, opened 2017-07-31
Fix #235 - support \ in use statements

### LINKED ISSUE #235: 'use' does not support namespaces
In investigating https://github.com/Microsoft/vscode/issues/30448, it seems that there is a bug with the grammar. I'm a bit confused since it seems like Atom Beta would have the most recent version of the grammar, but it doesn't actually repro in Atom Beta. When I link in language-php, I can actually repro it in Atom, so it must be on an older version? How do new versions of the grammar flow into Atom?

There are more repro cases in the linked issue, but the bug is basically this- have a 'use' statement inside a class, and that contains a `\`. Everything after it will have the `meta.use.php` scope.
```php
<?php
class Foo {
	use Some\Thing;

	// Comment
	public $name = 'foo';

	public function __construct() {
		$this->doStuff();
	}
}
```

(1) https://github.com/atom/language-php/blob/master/grammars/php.cson#L147 This `meta.use.php` section doesn't match the backslash in `Some\Thing`.

(2) https://github.com/atom/language-php/blob/master/grammars/php.cson#L174 This one is what matches, and only terminates after a `}` (which isn't applicable to this usage of `use`)

(3) https://github.com/atom/language-php/blob/master/grammars/php.cson#L1529 This is the one that should match.

I can help out with a PR, with some guidance as to the best way to fix it. Copy (3) into `class-body`? Change (1) to match `\`?

roblourens:
This fixes #235. It doesn't make use of `#class-name` as originally suggested, and maybe that would be a slightly better fix, but I couldn't get it to work - https://github.com/atom/language-php/issues/235#issuecomment-318920488. But this works and makes sense.

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -140,8 +140,8 @@
       {
         'match': '''(?xi)
           \\b(use)\\s+
-          ([a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*)
-          ((?:\\s*,\\s*[a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*)*)
+          ([a-z_\\x{7f}-\\x{ff}\\\\][a-z0-9_\\x{7f}-\\x{ff}\\\\]*)
+          ((?:\\s*,\\s*[a-z_\\x{7f}-\\x{ff}\\\\][a-z0-9_\\x{7f}-\\x{ff}\\\\]*)*)
           (?=\\s*;)
         '''
         'name': 'meta.use.php'
@@ -153,7 +153,7 @@
           '3':
             'patterns': [
               {
-                'match': '(?i)[a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*'
+                'match': '(?i)[a-z_\\x{7f}-\\x{ff}\\\\][a-z0-9_\\x{7f}-\\x{ff}\\\\]*'
                 'name': 'support.class.php'
               }
               {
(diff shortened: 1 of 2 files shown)


roblourens:
@50Wliu Have you had a chance to look at this yet? Would love to get the fix into the next vscode release, if we can merge it in the next couple days.

50Wliu:
Was planning to test this out yesterday, but had some unexpected npm issues while rebuilding Atom for (unrelated) changes.  First on my list though today.

50Wliu:
@roblourens I'm not seeing the spec failures that you saw when I switch the `support.class.php` matches to use `class-name`.  All of them pass for me.  Can you try out this diff for me (applied on top of this PR)?
```diff
diff --git a/grammars/php.cson b/grammars/php.cson
index 6f6ebc2..9f40189 100644
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -149,12 +149,15 @@
           '1':
             'name': 'keyword.other.use.php'
           '2':
-            'name': 'support.class.php'
+            'patterns': [
+              {
+                'include': '#class-name'
+              }
+            ]
           '3':
             'patterns': [
               {
-                'match': '(?i)[a-z_\\x{7f}-\\x{ff}\\\\][a-z0-9_\\x{7f}-\\x{ff}\\\\]*'
-                'name': 'support.class.php'
+                'include': '#class-name'
               }
               {
                 'match': ','
@@ -180,12 +183,15 @@
             '''
             'captures':
               '1':
-                'name': 'support.class.php'
+                'patterns': [
+                  {
+                    'include': '#class-name'
+                  }
+                ]
               '2':
                 'patterns': [
                   {
-                    'match': '(?i)[a-z_\\x{7f}-\\x{ff}\\\\][a-z0-9_\\x{7f}-\\x{ff}\\\\]*'
-                    'name': 'support.class.php'
+                    'include': '#class-name'
                   }
                   {
                     'match': ','
```

In addition, this PR will need some specs to ensure that #235 doesn't regress again.

roblourens:
Hm, it does work fine now, either #241 impacted it, since I rebased onto master right before submitting the PR, or I was doing something dumb before.

I'll fix that up and add specs.

50Wliu:
:+1:

PR #261, opened 2017-08-24
Improve number regexes

### LINKED ISSUE #260: binary integers not scoped
Hello, 

I came across this while doing some stuff with visual studio code, which led me here. PHP has a binary integer type, but the language-php file does not recognize this.

I did a quick and dirty add to my tmlanguage file but Im not sure if I should do a pullrequest for here.  So I might as well tell you what happens and what I did.
(also, this might be really wrong, I am sorry if it is Im not really proficient with regex)

i added this to patterns in interpolation
"match": "\\\\b[0-1]{1,2}",
"name": "constant.numeric.binary.php"

and added 
|(0b[0-1]*)
in the constant.numeric.php match expression.

Hope this helps! And I hope I did not do anything wrong by posing this here.

50Wliu:
These new regexes are taken mostly unmodified from language-javascript.  The only two changes are removing the optional `(o|O)` requirement for octal integers and renaming the period scope to be more inline with other punctuation scopes in language-php.  Other than fixing #260, this PR also fixes a different bug where floats like `.9` would be incorrectly tokenized as the string concatenation operator + the integer `9`.

Fixes #260

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -1385,14 +1385,6 @@
     # http://www.php.net/manual/en/language.types.string.php#language.types.string.parsing
     'patterns': [
       {
-        'match': '\\\\[0-7]{1,3}'
-        'name': 'constant.numeric.octal.php'
-      }
-      {
-        'match': '\\\\x[0-9A-Fa-f]{1,2}'
-        'name': 'constant.numeric.hex.php'
-      }
-      {
         'match': '\\\\[enrt\\\\\\$\\"]'
         'name': 'constant.character.escape.php'
       }
@@ -1907,6 +1899,9 @@
         'include': '#heredoc'
       }
       {
+        'include': '#numbers'
+      }
+      {
         'match': '(?i)\\bclone\\b'
         'name': 'keyword.other.clone.php'
       }
@@ -1977,9 +1972,6 @@
             'include': '#variable-name'
           }
         ]
-      }
-      {
-        'include': '#numbers'
       }
       {
         'include': '#instantiation'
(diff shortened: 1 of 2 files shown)


grammars/php.cson
```diff
@@ -1385,14 +1385,6 @@
     # http://www.php.net/manual/en/language.types.string.php#language.types.string.parsing
     'patterns': [
       {
-        'match': '\\\\[0-7]{1,3}'
```
Ingramz: Please revert these and rename their scopes to `constant.character.escape.octal.php` and `constant.character.escape.hex.php` respectively. While you are at it, please also consider adding the unicode escape highlighting that can be found from the link above.

grammars/php.cson
```diff
@@ -1385,14 +1385,6 @@
...
-        'name': 'constant.numeric.octal.php'
-      }
-      {
-        'match': '\\\\x[0-9A-Fa-f]{1,2}'
-        'name': 'constant.numeric.hex.php'
-      }
-      {
         'match': '\\\\[enrt\\\\\\$\\"]'
```
Ingramz: I think some ``\\`` can be pruned from here as they are in character class.

grammars/php.cson
50Wliu: Ah, so that's what these were.  I assumed that the other pattern would take care of them.  Will add them back and add some specs.

Ingramz: The appropriate table is located at http://ee1.php.net/manual/en/language.types.string.php#language.types.string.syntax.double

Ingramz:
Only thing that I am slightly concerened of is that PHP documentation wants to treat + and - as part of the integer literal, however it is allowed to have any amount of whitespace between the digits and the sign. This makes me believe that the current way of simply ignoring the sign (and letting those to be processed as infix operators) is better as we cannot handle edge cases in an elegant manner.

Decimal regexes can be slightly reduced in count.
Floating point (http://php.net/manual/en/language.types.float.php):
```
[0-9]*\\.[0-9]+
[0-9]+\\.[0-9]*
([0-9]+|[0-9]*\\.[0-9]+|[0-9]+\\.[0-9]*)[eE][+-]?[0-9]+
```
The last alternative is a bit verbose, at the beginning but it should it a little more obvious where the expression originates from.

Integers (http://php.net/manual/en/language.types.integer.php#language.types.integer.syntax):
```
0|[1-9][0-9]*
```

They can have scopes of their own this way, if needed.

50Wliu:
> This makes me believe that the current way of simply ignoring the sign (and letting those to be processed as infix operators) is better as we cannot handle edge cases in an elegant manner.

Yep.  We've tried treating `+/-` as part of the actual number multiple times in other languages (attempts that come to mind are JS, CSS, and Java) and every time it failed due to being unable to distinguish `3 - 5` from `- 5`.

PR #263, opened 2017-08-28
Improve number regexes

50Wliu:
Commit history got wiped due to a bad rebase (whoops!).  See #261 for the previous PR and review comments, which have been implemented in this PR.

/cc @Ingramz


50Wliu:
Hmm...the specs file got wiped.  Will need to check that back in.

grammars/php.cson
```diff
@@ -2091,6 +2091,7 @@
       {
         'match': '''(?x)
           (?<!\\$)(?:
```
Ingramz: What happens if one does not check for preceding/following dollar sign? Is there any kind of valid code that benefits from this check?

50Wliu: You're correct that no valid code benefits.  It does, however, prevent invalid code like `$0` from being tokenized as somewhat valid.

50Wliu:
I've decided to revert the decimal regexes back to language-javascript's implementation so that there's more consistency between languages.

No idea why that function spec is failing.

Ingramz:
I adapted a version of your previous pull request during weekend and now overlaid with this one (grammar only), has some new/different things:
https://github.com/atom/language-php/compare/fe28c0482463a1e153cacaaa4548da73934765c0...Ingramz:patch-1 ([patch](https://github.com/atom/language-php/compare/fe28c0482463a1e153cacaaa4548da73934765c0...Ingramz:patch-1.patch))

spec/php-spec.coffee
```diff
@@ -820,11 +820,11 @@ describe 'PHP grammar', ->
     it 'tokenizes function names with characters other than letters or numbers', ->
       # The space between foo and bar is a nbsp (char 160/hex 0xA0), not an actual space (char 32/hex 0x20)
       # 0xA0 is between 0x7F and 0xFF, making it a valid PHP identifier
-      tokens = grammar.tokenizeLines "<?php\nfunction foo bar() {}"
```
Ingramz: What happened here? This might be the reason for the tests failing

50Wliu: Wow, thanks for catching that.  I copied this file from my previous PR after it didn't make it through the rebase.  I guess Github's interface converts nbsps to spaces?

Ingramz:
@50Wliu I want to get this merged soon. Please consider the changes I proposed above in my last comment:
* Fix hexadecimal escape name (copy-paste issue)
* Specify base and type for numeric tokens, matching floats first makes it possible to match integers separately (if someone is interested in targeting different bases and floats).
* Remove boundary-checking. I just don't see where they are ever useful. If there is text preceding the digits, it will be parsed as something else, be it a variable, function name, class name, other constant. Text following a digit will simply deactivate the highlighting for the number, which arguably might make it harder to spot. PHP will still consider that part as a number, which would mean that getting rid of these boundaries will bring it more in line with PHP (same seems to apply to dollar sign). Also it will make the expressions look cleaner.

50Wliu:
I didn't see that comment.  I believe my last notification took me directly to the review and I didn't scroll up.  Thanks for the ping.

50Wliu:
@Ingramz I have implemented most of the changes in the patch.  In the sake of sanity, I haven't added the `integer` and `float` scopes so that I don't have to change all the numeric specs.

Ingramz:
@50Wliu LGTM 👍

PR #266, opened 2017-09-12
Always end PHP on ?>

### LINKED ISSUE #236: Incorrect color for switch default label when using alt-syntax
*Source: Microsoft/vscode#31320*

### Description
The keyword `default` is white like a regular label instead of purple if it is on a new line, has leading whitespace, and is followed by a colon.

This appears to be a follow-up to #226.

### Steps to reproduce
```php
switch ($a):
  default:
    break;
endswitch;
```

### LINKED ISSUE #246: Mixing HTML and PHP
This may be covered in an issue already but I couldn't find one that hits this exact case.

If you end the PHP block inside a function body, then the HTML after it is not colorized correctly.

```php
<div>foo</div>

<?php
foo(function() {
    ?>
    <div>foo</div>
    <?php
    return "sldkfj";
});
```

image

The orange `<div>foo</div>` is still `source.php` and `meta-function-call.php`. If the function line is commented out, or if the HTML isn't inside the function body, then it is handled correctly. Would this be expected to work?

### LINKED ISSUE #253: if/else alternative syntax colored incorrectly
https://github.com/Microsoft/vscode/issues/32142#issuecomment-321169718

PHP supports an alternative syntax for if/else that uses a colon instead of braces: http://php.net/manual/en/control-structures.alternative-syntax.php. In some cases, these are tokenized incorrectly as labels.

```php
<?php

$a = 5;

    if ($a == 5):
        echo "a equals 5";
        echo "...";
    elseif ($a == 6):
        echo "a equals 6";
        echo "!!!";
    else:
        echo "a is neither 5 nor 6";
    endif;
?>
```

When this block is indented, the `else` has `entity.name.goto-label.php`, while if, elseif, and endif are still correct.

If it's not indented, the `else` is tokenized correctly as well.

image

### LINKED ISSUE #273: Use statement with braces
From https://github.com/Microsoft/vscode/issues/34244#issuecomment-329987483

Not sure if this is related to https://github.com/atom/language-php/issues/246? `meta.use.php` persists unless there is a semicolon after the close brace.

```php
<?php

class Foo
{
    use A {
        B;
    }

    /**
     * Comment
     */
    private $abc;

		/**
		 * Comment 2
		 */
		 private $def;
}
```

image

50Wliu:
This PR restores the ability to intermingle HTML and PHP at the cost of losing scope context.  In addition, it fixes alternative if/else and switch/case highlighting.

Fixes #236
Fixes #246
Fixes #253
Fixes #273

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -164,7 +164,7 @@
             'beginCaptures':
               '0':
                 'name': 'punctuation.definition.use.begin.bracket.curly.php'
-            'end': '}'
+            'end': '}|(?=\\?>)'
             'endCaptures':
               '0':
                 'name': 'punctuation.definition.use.end.bracket.curly.php'
@@ -618,7 +618,7 @@
           '7':
             'name': 'punctuation.definition.array.begin.bracket.round.php'
         'contentName': 'meta.array.php'
-        'end': '\\)'
+        'end': '\\)|(?=\\?>)'
         'endCaptures':
           '0':
             'name': 'punctuation.definition.array.end.bracket.round.php'
@@ -799,7 +799,7 @@
             ]
           '2':
             'name': 'punctuation.definition.arguments.begin.bracket.round.php'
-        'end': '\\)'
+        'end': '\\)|(?=\\?>)'
         'endCaptures':
           '0':
             'name': 'punctuation.definition.arguments.end.bracket.round.php'
@@ -832,7 +832,7 @@
             ]
           '3':
             'name': 'punctuation.definition.arguments.begin.bracket.round.php'
-        'end': '\\)'
+        'end': '\\)|(?=\\?>)'
         'endCaptures':
           '0':
             'name': 'punctuation.definition.arguments.end.bracket.round.php'
@@ -1522,7 +1522,7 @@
             'beginCaptures':
               '0':
                 'name': 'punctuation.definition.namespace.begin.bracket.curly.php'
-            'end': '}'
+            'end': '}|(?=\\?>)'
             'endCaptures':
               '0':
                 'name': 'punctuation.definition.namespace.end.bracket.curly.php'
@@ -1553,7 +1553,7 @@
             'beginCaptures':
               '0':
                 'name': 'punctuation.definition.use.group.begin.bracket.curly.php'
-            'end': '}'
+            'end': '}|(?=\\?>)'
             'endCaptures':
               '0':
                 'name': 'punctuation.definition.use.group.end.bracket.curly.php'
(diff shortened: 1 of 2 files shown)


Ingramz:
I think the end patterns should be modified sparingly, there are a number of cases where one is not allowed to use end tag too early. This somewhat works to our benefit that we don't need to pollute the whole grammar with those. One example would be `use x\{a,b,c};` statements. 

There are two cases I can think of where things might break:
1) Where 'end' will not exit the whole stack when encountering `?>`
Example:
```php
<?php

namespace myNamespace {
  class myClass {
    function myFunction() {
      ?>
      Text to display
      <?php
    }
  }
}
```
2) Where a continuation will fail to highlight rest of the code as it doesn't know its scope anymore:
```php
<?php

switch ($variable) {
    case 'value':
        ?>
        Some text
        <?php
        break;

    default:
        # code...
        break;
}
```

Here the `default` token is not recognized as a switch case anymore.

Ingramz:
Case 1 example seems to be easily fixable, by adding the lookahead to its parent as an alternative.
https://github.com/atom/language-php/blob/542c7d9a27e3c75e7bba737437d17f5f9b2b66c5/grammars/php.cson#L1502

50Wliu:
1), like you said, is easily fixable by adding the lookahead to all the nested patterns.  2) of course is much harder, and I alluded to that in https://github.com/atom/language-php/issues/246#issuecomment-320261698.  I did however forget about that when making this PR.

Ingramz:
If the switch statement/goto-label is the only one for case 2, then it can most likely be worked out by excluding specific keywords inside the label rule directly, which makes the `default:` rule inside switch scope less necessary, as keywords aren't allowed to be used as labels anyway.

Unfortunately it is getting late here, so I will try reviewing everything first thing in the morning.

Ingramz:
@50Wliu I tried getting rid of the ones that I believe to be unnecessary. Please take a look and let me know if you disagree with any of those: https://github.com/atom/language-php/compare/wl-end-php...Ingramz:patch-2

In the meanwhile, I am trying to find any poorly highlighted continuations for the remaining ones.

Ingramz:
Also fixes #253

50Wliu:
Thanks, updated.

ungb:
I was able to test this out, by testing 1.15.0 and the changes here.  

I tried going through each of the issues to test each scenario.  After following up with @50Wliu, everything looks like it's working correctly.

(1.15.0 | changes from this PR | 1.20 stable with no packages)
https://github.com/atom/language-php/issues/236
image


https://github.com/atom/language-php/issues/246

image

https://github.com/atom/language-php/issues/253
image


I didn't see any issues with this changes looking at other php files, but someone else might know better who use php day in and day out.

roblourens:
I also spent some time trying cases from the referenced issues and clicking through a PHP project, and it all looks great!

Ben3eeE:
:shipit:

PR #269, opened 2017-09-13
Bring in suggested changes

50Wliu:
Let's see if this create PR button works.


grammars/php.cson
```diff
@@ -1679,6 +1651,7 @@
       }
       {
         'match': '''(?x)
+          (^\\s*)?
```
50Wliu: Normally I would be against including something like this because it effectively moves this pattern to the top of the stack but given the circumstances I think it's ok.

spec/php-spec.coffee
```diff
@@ -820,11 +820,12 @@ describe 'PHP grammar', ->
     it 'tokenizes function names with characters other than letters or numbers', ->
       # The space between foo and bar is a nbsp (char 160/hex 0xA0), not an actual space (char 32/hex 0x20)
       # 0xA0 is between 0x7F and 0xFF, making it a valid PHP identifier
-      tokens = grammar.tokenizeLines "<?php\nfunction foo bar() {}"
+      functionName = "foo#{String.fromCharCode 160}bar"
```
50Wliu: :zap:
You can probably condense the above comment to just say that 160 is allowed because its hex representation is between 0x7f-0xff.

PR #274, opened 2017-09-16
End complex use statements properly

### LINKED ISSUE #273: Use statement with braces

50Wliu:
### Requirements

### Description of the Change

This is a very embarrassing regression caused by #258 that I didn't notice because the specs were testing _only_ the use patterns and not that they were terminated correctly.

### Applicable Issues

Fixes #273

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -142,7 +142,7 @@
         'beginCaptures':
           '0':
             'name': 'keyword.other.use.php'
-        'end': '(?=;)'
+        'end': '(?<=})|(?=;)'
         'name': 'meta.use.php'
         'patterns': [
           {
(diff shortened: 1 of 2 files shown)


Ingramz:
It's been already dealt with in #266

50Wliu:
Really, I tried with #266 and I think it was still broken.  Let me double check.

50Wliu:
Ah, you're right!  Forgot to pull...:sweat:

PR #275, opened 2017-09-18
Correct __toString capitalisation

tpunt:
### Description of the Change

Correct the capitalisation of the `__toString` magic method. The other magic methods follow the camelCase convention, so this one should too.

E.g:
incorrect __toString capitalisation

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -1782,7 +1782,7 @@
           ((?:(?:final|abstract|public|private|protected|static)\\s+)*)
           (function)\\s+
           (?i:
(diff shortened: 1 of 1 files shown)


Ingramz:
I approve the change to maintain consistency, however it should make little difference functionality wise to the current (v0.40.0 and up) grammar as that part is case insensitive.

@tpunt are you running an older version of Atom or language-php package by chance?

tpunt:
@Ingramz It would appear so. Looks like I should start upgrading my dev environment... Thanks anyway.

Issue #221, opened 2017-06-21
Static properties and functions are not styled


jens1o:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

![image](https://user-images.githubusercontent.com/11234139/27384034-aed33b2a-568d-11e7-8339-3ab6ae8ef752.png)

### Steps to Reproduce

Write php code:
```php
<?php

Sapi::sendResponse($response);
```

**Expected behavior:** [What you expect to happen]
Styled like it was before the tremendous clean up

**Actual behavior:** [What actually happens]
see screenshot
**Reproduces how often:** [What percentage of the time does it reproduce?]
100%
### Versions

I'm using vscode.

jens1o:
This also affects the class annotation:
![image](https://user-images.githubusercontent.com/11234139/27384400-fdf34f78-568e-11e7-9d96-eb174f59b9c6.png)

50Wliu:
Hmm...it's working for me, and we also have specs covering it - can you make sure your syntax theme highlights `support.class` and `keyword.operator.class`?

![php-static](https://user-images.githubusercontent.com/2766036/27417240-f6f5f110-56e0-11e7-82dd-9bef351e50fc.png)

jens1o:
Then I think it's a vscode bug? 
![image](https://user-images.githubusercontent.com/11234139/27433151-b5b3a71a-5753-11e7-882f-2d50b059f8c9.png)

50Wliu:
Yeah, it might be.  We have a lot of specs testing it [here](https://github.com/atom/language-php/blob/e1290265f3d68316347e0ab2665686016b4b24b7/spec/php-spec.coffee#L616).

I would bring it up with the VSCode developers and see what they say :).

50Wliu:
[...]

PR #280, opened 2017-09-23
Fix issues with TextMate compatibility

50Wliu:
### Requirements

### Description of the Change

atom/first-mate#100 tweaked first-mate to be more compliant with the original TextMate grammar implementation.  Rules in language-php that were relying on that extra newline have been rewritten to match even if there is no trailing newline.

### Alternate Designs

Alternative first-mate fix.

### Benefits

language-php will work with first-mate@7.0.8+.  This _might_ fix VSCode as well, though I have no idea how I would go about testing that :).

### Possible Drawbacks

These changes are mostly cosmetic.  The previous end rules were "ensure that there is a next character that is not an identifier", now they're "ensure that the next character is not an identifier".  Subtle difference :).

### Applicable Issues

#221?

/cc @Ingramz
/cc @jens1o @roblourens @aeschli @alexandrudima for VSCode testing

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -207,7 +207,7 @@
     'patterns': [
       {
         'begin': '(?i)(?=\\\\?[a-z_0-9]+\\\\)'
-        'end': '(?i)([a-z_][a-z_0-9]*)?(?=[^a-z0-9_\\\\])'
+        'end': '(?i)([a-z_][a-z_0-9]*)?(?![a-z0-9_\\\\])'
         'endCaptures':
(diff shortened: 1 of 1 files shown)


jens1o:
@aeschli @alexandrudima Is there a safe way I can test it, since Insiders releases are not available due to hurricane Maria?

50Wliu:
I am merging this so that we can get the first-mate fixes included in Atom.  Hope everyone's safe after Hurricane Maria!

PR #285, opened 2017-10-05
Add support for ternary and null coalescing operators

dsifford:
This PR adds scopes for ternaries and the new (as of php7.0) null-coalescing operator.

### Description of the Change

Currently, this is how the above items are scoped:

```php
// ternary expression
  $foo = $bar ? $bar : 'baz';
//            ^text.html.php meta.embedded.block.php source.php
//                   ^text.html.php meta.embedded.block.php source.php

// ternary shorthand
  $foo = $bar ?: 'baz';
//            ^^text.html.php meta.embedded.block.php source.php

// null coalescing operator
  $foo = $bar ?? 'baz';
//            ^^text.html.php meta.embedded.block.php source.php
```

And here's how it is after the changes I made:

```php
// ternary expression
  $foo = $bar ? $bar : 'baz';
//            ^text.html.php meta.embedded.block.php source.php keyword.operator.ternary.php
//                   ^text.html.php meta.embedded.block.php source.php keyword.operator.ternary.php

// ternary shorthand
  $foo = $bar ?: 'baz';
//            ^^text.html.php meta.embedded.block.php source.php keyword.operator.ternary.php

// null coalescing operator
  $foo = $bar ?? 'baz';
//            ^^text.html.php meta.embedded.block.php source.php keyword.operator.null-coalescing.php
```

### Alternate Designs

It would be possible to expand the specificity of the ternary expression scope (e.g. `(expr) : (val_if_true) ? (val_if_false)`) by wrapping the entire expression in `meta.expression.ternary.php` (or something along those lines), but I chose not to do that because it doesn't seem like it would add any value.

### Benefits

- Increased scope specificity.
- Allows theme maintainers to colorize ternary and null coalescing operators the same as the other existing operators.

### Possible Drawbacks

None to my knowledge.

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -2041,6 +2041,15 @@
         'match': ','
         'name': 'punctuation.separator.delimiter.php'
       }
+      {
+        'include': '#ternary_expression'
+      }
+      {
+        'include': '#ternary_shorthand'
+      }
+      {
+        'include': '#null_coalescing'
+      }
     ]
   'namespace':
     'begin': '(?i)(?:(namespace)|[a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*)?(\\\\)'
@@ -3695,6 +3704,7 @@
       }
     ]
   'variables':
+  
     'patterns': [
       {
         'include': '#var_language'
@@ -3724,3 +3734,32 @@
         ]
       }
     ]
+  
+  'ternary_shorthand':
+    'match': '\\s*(\\?:)\\s*'
+    'captures':
+      '1':
+        'name': 'keyword.operator.ternary.php'
+
+  'ternary_expression':
+    'begin': '(?<!\\?)\\s*(\\?)\\s*(?!:|\\?)'
+    'beginCaptures':
+      '1':
+        'name': 'keyword.operator.ternary.php'
+    'end': '\\s*(:)\\s*'
+    'endCaptures':
+      '1':
+        'name': 'keyword.operator.ternary.php'
+    'patterns': [
+      {
+        'include': '#language'
+      }
+      {
+        'include': '$self'
+      }
+    ]
+  'null_coalescing':
+    'match': '\\s*(\\?\\?)\\s*'
+    'captures':
+      '1':
+        'name': 'keyword.operator.null-coalescing.php'
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -3704,7 +3704,6 @@
       }
     ]
   'variables':
-  
     'patterns': [
       {
         'include': '#var_language'
@@ -3734,13 +3733,9 @@
         ]
       }
     ]
-  
   'ternary_shorthand':
-    'match': '\\s*(\\?:)\\s*'
-    'captures':
-      '1':
-        'name': 'keyword.operator.ternary.php'
-
+    'match': '\\?:'
+    'name': 'keyword.operator.ternary.php'
   'ternary_expression':
     'begin': '(?<!\\?)\\s*(\\?)\\s*(?!:|\\?)'
     'beginCaptures':
@@ -3759,7 +3754,5 @@
       }
     ]
   'null_coalescing':
-    'match': '\\s*(\\?\\?)\\s*'
-    'captures':
-      '1':
-        'name': 'keyword.operator.null-coalescing.php'
+    'match': '\\?\\?'
+    'name': 'keyword.operator.null-coalescing.php'
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -3741,9 +3741,9 @@
     'beginCaptures':
       '1':
         'name': 'keyword.operator.ternary.php'
-    'end': '\\s*(:)\\s*'
+    'end': ':'
     'endCaptures':
-      '1':
+      '0':
         'name': 'keyword.operator.ternary.php'
     'patterns': [
       {
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -2042,13 +2042,13 @@
         'name': 'punctuation.separator.delimiter.php'
       }
       {
-        'include': '#ternary_expression'
-      }
-      {
         'include': '#ternary_shorthand'
       }
       {
         'include': '#null_coalescing'
+      }
+      {
+        'include': '#ternary_expression'
       }
     ]
   'namespace':
@@ -3737,9 +3737,9 @@
     'match': '\\?:'
     'name': 'keyword.operator.ternary.php'
   'ternary_expression':
-    'begin': '(?<!\\?)\\s*(\\?)\\s*(?!:|\\?)'
+    'begin': '\\?'
     'beginCaptures':
-      '1':
+      '0':
         'name': 'keyword.operator.ternary.php'
     'end': ':'
     'endCaptures':
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -3749,9 +3749,6 @@
       {
         'include': '#language'
       }
-      {
-        'include': '$self'
-      }
     ]
   'null_coalescing':
     'match': '\\?\\?'
--- a/spec/php-spec.coffee
+++ b/spec/php-spec.coffee
@@ -244,7 +244,7 @@ describe 'PHP grammar', ->
           """
           expect(tokens[1][7]).toEqual value: '?:', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'keyword.operator.ternary.php']
           expect(tokens[1][11]).toEqual tokens[1][7]
-          expect(tokens[1][7]).toEqual tokens[1][15]
+          expect(tokens[1][15]).toEqual tokens[1][7]
 
         it 'should tokenize a combination of ternaries', ->
           tokens = grammar.tokenizeLines """<?php
--- a/spec/php-spec.coffee
+++ b/spec/php-spec.coffee
@@ -243,7 +243,7 @@ describe 'PHP grammar', ->
             $foo = false ?: false ?: true ?: false;
           """
           expect(tokens[1][7]).toEqual value: '?:', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'keyword.operator.ternary.php']
-          expect(tokens[1][7]).toEqual tokens[1][11]
+          expect(tokens[1][11]).toEqual tokens[1][7]
           expect(tokens[1][7]).toEqual tokens[1][15]
 
         it 'should tokenize a combination of ternaries', ->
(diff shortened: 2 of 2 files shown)


grammars/php.cson
```diff
@@ -3695,6 +3704,7 @@
       }
     ]
   'variables':
+
```
50Wliu: :fire: empty newlines

grammars/php.cson
```diff
@@ -3724,3 +3734,32 @@
         ]
       }
     ]
+  
+  'ternary_shorthand':
+    'match': '\\s*(\\?:)\\s*'
```
50Wliu: Don't capture the spaces, they're not adding anything of value.  Then this can be shortened to `'match': '\\?:'`, and you can scope the 0th capture.

grammars/php.cson
```diff
@@ -3724,3 +3734,32 @@
...
+      }
+      {
+        'include': '$self'
+      }
+    ]
+  'null_coalescing':
+    'match': '\\s*(\\?\\?)\\s*'
+    'captures':
+      '1':
+        'name': 'keyword.operator.null-coalescing.php'
```
50Wliu: Add a newline after the last line

grammars/php.cson
```diff
@@ -3724,3 +3734,32 @@
...
+    'patterns': [
+      {
+        'include': '#language'
+      }
+      {
+        'include': '$self'
+      }
+    ]
+  'null_coalescing':
+    'match': '\\s*(\\?\\?)\\s*'
```
50Wliu: Same comment here: spaces are unnecessary

grammars/php.cson
```diff
@@ -3724,3 +3734,32 @@
...
     ]
+  
+  'ternary_shorthand':
+    'match': '\\s*(\\?:)\\s*'
+    'captures':
+      '1':
+        'name': 'keyword.operator.ternary.php'
+
+  'ternary_expression':
+    'begin': '(?<!\\?)\\s*(\\?)\\s*(?!:|\\?)'
```
50Wliu: I think this can be simplified to a simple `\\?` if you include `#ternary_shorthand` and `#null_coalescing` _before_ `ternary_expression`.

spec/php-spec.coffee
```diff
@@ -209,6 +209,53 @@ describe 'PHP grammar', ->
...
+          """
+          expect(tokens[1][6]).toEqual value: '?', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'keyword.operator.ternary.php']
+          expect(tokens[1][8]).toEqual value: ':', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'keyword.operator.ternary.php']
+        
+        it 'should tokenize shorthand ternaries', ->
+          tokens = grammar.tokenizeLines """<?php
+            $foo = false ?: false ?: true ?: false;
+          """
+          expect(tokens[1][7]).toEqual value: '?:', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'keyword.operator.ternary.php']
+          expect(tokens[1][7]).toEqual tokens[1][11]
```
50Wliu: I would switch these around to be `expect(tokens[1][11]).toEqual tokens[1][7]` to make it more clear what we're asserting.

grammars/php.cson
dsifford: Would that create a more fragile situation though? Thought about that as well but I went this route to keep it future proof.

grammars/php.cson
dsifford: Used to being spoiled by editorconfig. Good catch

grammars/php.cson
dsifford: Wasn't sure if textmate ate spaces or you had to be explicit with each token. 

To clarify: Textmate _does_ eat spaces when not defined explicitly?

grammars/php.cson
```diff
@@ -2041,6 +2041,15 @@
         'match': ','
         'name': 'punctuation.separator.delimiter.php'
       }
+      {
```
Ingramz: Is there any particular reason for the includes? Could those simply be just inlined?

dsifford: Personal preference. Never know when they might be needed to compose some larger expression in the future. Ran into that problem a couple times with inlined matches in `language-css`.

grammars/php.cson
50Wliu: I'm not sure what you mean by Textmate "eat"ing spaces.  But in both cases the spaces won't have specific tokens associated with them.

grammars/php.cson
50Wliu: @Ingramz what do you think about this?

grammars/php.cson
dsifford: To clarify what I meant by "eating"... 

An example of a simple grammar in extended Backus-Naur form could be this...

```
letter ::= "A" | "B" ;
number ::= "0" | "1" ;
identifier ::= letter , { letter | number };
program ::= { identifier };
```

Here, `ABBA10AB` would be a valid grammar, but not `ABBA10 AB` because whitespace must be explicitly accounted for (i.e., whitespace is not "eaten").

So, to get the above to work for the second form with spaces, you'd have to have the grammar something like this...

```
letter ::= "A" | "B" ;
number ::= "0" | "1" ;
whitespace ::= " " | "\t" | "\n" | "\r" ;
identifier ::= letter , { letter | number };
program ::= { identifier, whitespace };
```

.....

My point: Does TextMate "eat" whitespace is asking whether or not a grammar will lock up if it hits a whitespace character not explicitly accounted for.

dsifford: Done. ✅

grammars/php.cson
dsifford: Done. ✅

grammars/php.cson
dsifford: Done. ✅

grammars/php.cson
dsifford: Done. ✅

spec/php-spec.coffee
dsifford: Done ✅

grammars/php.cson
50Wliu: Not everything has to have scopes.  If it hits a character that doesn't match anything, it ignores it and moves onto the next character.

grammars/php.cson
```diff
@@ -3724,3 +3733,26 @@
...
     ]
+  'ternary_shorthand':
+    'match': '\\?:'
+    'name': 'keyword.operator.ternary.php'
+  'ternary_expression':
+    'begin': '(?<!\\?)\\s*(\\?)\\s*(?!:|\\?)'
+    'beginCaptures':
+      '1':
+        'name': 'keyword.operator.ternary.php'
+    'end': '\\s*(:)\\s*'
```
50Wliu: Same issue with spaces

grammars/php.cson
```diff
@@ -3724,3 +3733,26 @@
...
+    'endCaptures':
+      '1':
+        'name': 'keyword.operator.ternary.php'
+    'patterns': [
+      {
+        'include': '#language'
+      }
+      {
+        'include': '$self'
```
50Wliu: What's this `$self` match doing here?

spec/php-spec.coffee
```diff
@@ -209,6 +209,53 @@ describe 'PHP grammar', ->
...
+          expect(tokens[1][6]).toEqual value: '?', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'keyword.operator.ternary.php']
+          expect(tokens[1][8]).toEqual value: ':', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'keyword.operator.ternary.php']
+        
+        it 'should tokenize shorthand ternaries', ->
+          tokens = grammar.tokenizeLines """<?php
+            $foo = false ?: false ?: true ?: false;
+          """
+          expect(tokens[1][7]).toEqual value: '?:', scopes: ['text.html.php', 'meta.embedded.block.php', 'source.php', 'keyword.operator.ternary.php']
+          expect(tokens[1][11]).toEqual tokens[1][7]
+          expect(tokens[1][7]).toEqual tokens[1][15]
```
50Wliu: Switch this as well

grammars/php.cson
dsifford: Aha... Makes sense. Thanks for clarifying.

grammars/php.cson
dsifford: Inside a ternary can be any valid php expression. Since there is no "expression" defined, I just reinclude the entire language so that everything is available. 

Is this not to correct way to do that?

spec/php-spec.coffee
dsifford: Whoops! Done ✅

grammars/php.cson
dsifford: Done ✅

grammars/php.cson
dsifford: Oh whoops, I misstated that.. I think I put the self there along with language because ternaries can be nested.

50Wliu: Only `#language` should be included as `$self` would include the HTML grammar instead: https://github.com/atom/language-php/blob/aa095129f29f4ffbf5db6f154aa283997ddbca29/grammars/php.cson#L135

dsifford: Ah.. I think my confusion was that I thought `$self` was scoped at or below the current scope level and not the entire grammar. 

Thanks for the clarification.

50Wliu: `$self` includes everything in the top-level `patterns` block.

dsifford: Got it. Fixed.

grammars/php.cson
dsifford: @50Wliu To my knowledge, this is that last thing that needs addressing with this PR, correct?

50Wliu: Yep, I'll bring it up with him.

Ingramz: I'm sorry, it got buried by GitHub marking it "outdated".

I vote for reordering these rules. It will make the rules more fragile if they are viewed independently from eachother and if you wanted to include them selectively, but if they are grouped together (which is currently the case and most likely will stay this way), the two other rules do the same job of asserting the negative lookahead/lookbehind of the third rule without the need of repeating ourselves. It's as efficient or better in terms of matching and should be easier to read for anyone seeing this for the first time.

dsifford: Works for me! I'll adjust later on this afternoon 👍 

Re: GitHub marking outdated.... I can relate. That drives me nuts!

dsifford: @50Wliu All set now. To my knowledge I've now addressed everything that was brought up.

PR #292, opened 2017-10-21
Additional changes for the split grammars

Ingramz:
cc @50Wliu

### DIFF
--- a/grammars/html.cson
+++ b/grammars/html.cson
@@ -35,6 +35,9 @@
       (?:php|phtml)
     (?=\\s|:|$)
   )
+  |
+  # Regular opening PHP tags
+  ^\\s*<\\?(?i:php|=|\\s|$)
 '''
 'foldingStartMarker': '(/\\*|\\{\\s*$|<<<HTML)'
 'foldingStopMarker': '(\\*/|^\\s*\\}|^HTML;)'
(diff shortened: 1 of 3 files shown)


grammars/html.cson
```diff
@@ -35,6 +35,9 @@
       (?:php|phtml)
     (?=\\s|:|$)
   )
+  |
+  # Regular opening PHP tags
+  ^\\s*<\\?
```
50Wliu: This would match the XML tag too, no?

Ingramz: Yes it would. I don't want to mess with grammar scoring right now, so I will revert that to regular PHP tag and short echo tag instead.

spec/html-spec.coffee
```diff
@@ -73,17 +73,54 @@ describe 'PHP in HTML', ->
...
+      expect(lines[1][0]).toEqual value: '<?php', scopes: ['text.html.php', 'meta.embedded.line.php', 'punctuation.section.embedded.begin.php']
+
+    it 'does not recognize shebang on any of the other lines', ->
+      lines = grammar.tokenizeLines '''
+
+        #!/usr/bin/env php
+        <?php echo "test"; ?>
+      '''
+
+      expect(lines[1][0]).toEqual value: '#!/usr/bin/env php', scopes: [ 'text.html.php' ]
```
50Wliu: Remove array padding

50Wliu:
Needs a rebase.  Sorry about that.

PR #302, opened 2017-11-29
Handle quirk in how PHP parses identifiers

### LINKED ISSUE #301: Arabic variable names not highlighted
```php
<?php

$abc = "a";
$سن = "a";

```

image

This is legal in PHP apparently.

https://github.com/Microsoft/vscode/issues/34258

50Wliu:
### Requirements

### Description of the Change

While the PHP manual documents the range of valid identifiers as `[a-z_\x{7f}-\x{ff}][a-z0-9_\x{7f}-\x{ff}]*`, due to a quirk with how PHP parses identifiers it looks at each _byte_ rather than _characters_.  Therefore while `سن` is technically not in that range, its hex representation is `d8 b3 d9 86`, and PHP recognizes it as a valid identifier. To handle this behavior in Atom, we use the modified regex `[a-z_\x{7f}-\x{7fffffff}][a-z0-9_\x{7f}-\x{7fffffff}]*`.  Thanks to @Ingramz for the fix.

### Alternate Designs

Just don't fix this as it's really an odd quirk of PHP.

### Benefits

Arabic characters will be tokenized as identifiers where valid.

### Possible Drawbacks

Unknown.

### Applicable Issues

Fixes #301

Specs coming later.

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -1,11 +1,18 @@
 # Actual PHP grammar
+
+# NOTE: While the PHP manual documents the range of valid identifiers as [a-z_\x{7f}-\x{ff}][a-z0-9_\x{7f}-\x{ff}]*,
+# due to a quirk with how PHP parses identifiers it looks at each _byte_ rather than _characters_.
+# Therefore while سن is technically not in that range, its hex representation is `d8 b3 d9 86`,
+# and PHP recognizes it as a valid identifier. To handle this behavior in Atom, we use the modified regex
+# [a-z_\x{7f}-\x{7fffffff}][a-z0-9_\x{7f}-\x{7fffffff}]*.
+# See https://github.com/atom/language-php/issues/301.
 'scopeName': 'source.php'
 'patterns': [
   {
     'include': '#comments'
   }
   {
-    'begin': '(?i)^\\s*(interface)\\s+([a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*)\\s*(extends)?\\s*'
+    'begin': '(?i)^\\s*(interface)\\s+([a-z_\\x{7f}-\\x{7fffffff}][a-z0-9_\\x{7f}-\\x{7fffffff}]*)\\s*(extends)?\\s*'
     'beginCaptures':
       '1':
         'name': 'storage.type.interface.php'
@@ -13,12 +20,12 @@
         'name': 'entity.name.type.interface.php'
       '3':
         'name': 'storage.modifier.extends.php'
-    'end': '(?i)((?:[a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*\\s*,\\s*)*)([a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*)?\\s*(?:(?={)|$)'
+    'end': '(?i)((?:[a-z_\\x{7f}-\\x{7fffffff}][a-z0-9_\\x{7f}-\\x{7fffffff}]*\\s*,\\s*)*)([a-z_\\x{7f}-\\x{7fffffff}][a-z0-9_\\x{7f}-\\x{7fffffff}]*)?\\s*(?:(?={)|$)'
     'endCaptures':
       '1':
         'patterns': [
           {
-            'match': '(?i)[a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*'
+            'match': '(?i)[a-z_\\x{7f}-\\x{7fffffff}][a-z0-9_\\x{7f}-\\x{7fffffff}]*'
             'name': 'entity.other.inherited-class.php'
           }
           {
@@ -36,7 +43,7 @@
     ]
   }
   {
-    'begin': '(?i)^\\s*(trait)\\s+([a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*)'
+    'begin': '(?i)^\\s*(trait)\\s+([a-z_\\x{7f}-\\x{7fffffff}][a-z0-9_\\x{7f}-\\x{7fffffff}]*)'
     'beginCaptures':
       '1':
         'name': 'storage.type.trait.php'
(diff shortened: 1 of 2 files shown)


Ingramz:
To clarify, why unicode code points work without the unicode support in PHP, is quite simple: UTF-8 encodes all unicode code points past 0x7F in bytes from range 0x80-0xFF, leaving the original ASCII part (0x00-0x7F, defined same way in all locales) intact.

Whoever decided that PHP parser should still match characters 0x7F-0xFF was quite smart to do so or got lucky. Regardless of which character we mean by a specific byte or byte sequence from range 0x80-0xFF (usually defined by locale), PHP simply consumes them. This means that if we try matching `ä` in ISO-8859-1 encoding (`e4`) or UTF-8 encoding (`c3 a4`), it wouldn't make much of a difference to PHP as in both cases they are bytes from range 0x80-0xFF.

The regular expression, found from PHP manual, `[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*` is then correct if matching is done per byte.

Where we went wrong is that the text we match using regular expressions is in Unicode, meaning that we are not presented a byte sequence `c3 a4`, as that is automatically converted to `e4` (U+00E4). When that idea is extended to code points greater than 0xFF, then in case of for instance U+2620 (☠️), PHP sees 3 bytes `e2 98 a0` but Oniguruma sees one code point `\x2620`, that cannot be matched using `\xe2\x98\xa0` (`\x{2620}` has to be used instead). Notice that the 3 byte version uses bytes from range 0x80-0xFF.

Because unicode code points that do not belong to ASCII are always constructed from bytes belonging to range 0x80-0xFF, we can match these arbitrary length byte sequences by matching also all of the unicode code points. Per documentation, Oniguruma's highest code point is `7fffffff`, something that UTF-8 cannot encode, but if it ever becomes an issue, we can fix it by adjusting the range to a lower value.

Now you might ask, what about UTF-16 or UTF-32 or any other obscure encoding, then I have some good news. When [zend.multibyte](http://php.net/manual/en/ini.core.php#ini.zend.multibyte) is enabled (prerequisite for such encodings), you can still use unicode code points in identifiers. If that ever changes, we might need to rethink our strategy, but considering this is almost always turned off and disabling unicode is undesired, which is also why we don't have to worry about this anytime soon.

It should be noted though, just because you can use emojis or other unicode code points in your variable and function names, doesn't mean you should. There is some [locale dependent downcasing being done](https://stackoverflow.com/a/3422946), which can result in the code not being portable. It's still a neat party trick though. (Or a way to localize variables/function names for educational use).

grammars/php.cson
```diff
@@ -3493,7 +3500,7 @@
         'captures':
           '1':
             'name': 'punctuation.definition.variable.php'
-        'match': '(?i)(\\$+)[a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*\\b'
+        'match': '(?i)(\\$+)[a-z_\\x{7f}-\\x{7fffffff}][a-z0-9_\\x{7f}-\\x{7fffffff}]*\\b'
```
Ingramz: Word boundary needs to be removed from here as a lot of unicode code points are not considered as word characters, simplest case are variables such as `$🐘` This might be also needed to be done elsewhere, I'll take a look at the other changed lines later.

Since we are pretty much matching all of the characters in existence, except for the few in ASCII range, there is no need for a boundary check as it terminates at the characters not specified here.

grammars/php.cson
```diff
@@ -141,7 +148,7 @@
             'match': '''(?xi)
               \\b(as)
               \\s+(final|abstract|public|private|protected|static)
-              \\s+([a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*)
+              \\s+([a-z_\\x{7f}-\\x{7fffffff}][a-z0-9_\\x{7f}-\\x{7fffffff}]*)
               \\b
```
Ingramz: Removing end boundary should cause no harm.

grammars/php.cson
```diff
@@ -155,7 +162,7 @@
           {
             'match': '''(?xi)
               \\b(as)
-              \\s+([a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*)
+              \\s+([a-z_\\x{7f}-\\x{7fffffff}][a-z0-9_\\x{7f}-\\x{7fffffff}]*)
               \\b
```
Ingramz: Removing end boundary should cause no harm.

grammars/php.cson
```diff
@@ -1299,7 +1306,7 @@
       }
       {
         # Root namespace function calls (built-in or user)
-        'begin': '(?i)(\\\\)?\\b([a-z_\\x{7f}-\\x{ff}][a-z0-9_\\x{7f}-\\x{ff}]*)\\s*(\\()'
+        'begin': '(?i)(\\\\)?\\b([a-z_\\x{7f}-\\x{7fffffff}][a-z0-9_\\x{7f}-\\x{7fffffff}]*)\\s*(\\()'
```
Ingramz: Same as above, `(?<![a-z0-9_\\x{7f}-\\x{7fffffff}])` should be a good replacement.

grammars/php.cson
```diff
@@ -1269,8 +1276,8 @@
         'begin': '''(?xi)
           (
             \\\\?\\b                                                # Optional root namespace
```
Ingramz: Without much worrying about it, `(?<![a-z0-9_\\x{7f}-\\x{7fffffff}])` should be sufficient replacement for `\\b`. If someone can prove, it could potentially be removed or replaced by something simpler.

50Wliu: Would `(?<!\\d)` also work?

Ingramz: It is supposed to assert that none of the characters usable in identifier come immediately before the start of the pattern. If we want to retain the meaning of the word boundary there, then I believe we cannot use a subset of the pattern I proposed or else it might match in some strange cases.

I don't know how to efficiently prove that any of the subsets would work. One would need to know all the possible patterns that can precede this pattern and then verify that none of them can match the characters at the end which you are trying to omit. Then also the same for cases where no match could have been made and the cursor was simply advanced by one. It is easier to just match a wider range of characters as it definitely will not do any harm.

P.S. In unicode mode of oniguruma `\d` seems to match also other characters besides `[0-9]` outside ASCII, so one should be careful when using it.

PR #307, opened 2018-01-24
Added firstlineMatch support to language-php

Skeeve:
so that we do not have to rely on the filename's extension for recognizing a php file.

### Description of the Change

Added firstlineMatch support to language-php so that we do not have to rely on the filename's extension for recognizing a php file.

### Alternate Designs

I cannot think of another design for this except maybe falling back to the "file" command which would be a problem on non-IX systems.  

### Benefits

PHP files are recognized not only by extension but also by content.

### Possible Drawbacks

Due to my limited (just starting with atom and php) knowledge, the firstlineMatch might not be perfect.

### Applicable Issues

Not an issue but a [conversation](https://discuss.atom.io/t/better-language-recognition-please/51795).

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -5,24 +5,20 @@
 
 ### Description of the Change
 
-<!--
-
-We must be able to understand the design of your change from this description. If we can't get a good idea of what the code will be doing from the description here, the pull request may be closed at the maintainers' discretion. Keep in mind that the maintainer reviewing this PR may not be familiar with or have worked with the code here recently, so please walk us through the concepts.
-
--->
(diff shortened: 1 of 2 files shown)


50Wliu:
We already do; it's in `html.cson` which wraps `php.cson`.
https://github.com/atom/language-php/blob/master/grammars/html.cson

I just tested with a `de.lang` file that started with `<?php` and it automatically set the language to PHP.  If that is not happening to you, let's bring it back to the Discuss topic and start by saying what version of Atom you're using and whether or not it reproduces in safe mode (`atom --safe`).

Skeeve:
Okay. Will check. As I installed atom just 2 days ago, it's a fairly new version.

Issue #130, opened 2016-02-29
Switch statement


mastazi:
In a php `switch` statement, the lines immediately following a `:` are supposed to be indented (in this case, the line with the `print` statement and the line with the `break` statement should be indented with respect to the line with the `case` statement):

    <?php
    $myvar = 2
    switch($myvar){
      case 1:
        print 'One';
        break;
    }

but instead, in Atom, they are not indented:

    <?php
    $myvar = 2
    switch($myvar){
      case 1:
      print 'One';
      break;
    }

The same issue happens when pasting code (which contains a switch statement) from an external source.

Reproduced in safe mode with updated Atom and searched for existing issues as follows:
`is:issue user:atom -repo:atom/electron "switch statement" `

Arcanemagus:
This is a (better written) duplicate of #30.

mostafamabrouk:
Any Update on this issue ?

Arcanemagus:
Just to confirm this as `triaged`, I've reproduced on Atom v1.40.0-nightly1.

Arcanemagus:
Adding some information from #294:

> If the case is a string that contains `if` indentation is counted as if there was an `if` statement on that line, leading to it breaking in other manners.

PR #313, opened 2018-02-23
Indent lines immediately following switch case keyword

stianlik:
### Description of the Change

Add the `case` keyword to `increaseIndentPattern`  to ensure that the line  immediately following switch case keyword is indented properly, see issue #130 .

### Alternate Designs

No alternate designs.

### Benefits

Better code indentation when writing switch statements.

### Possible Drawbacks

Should not be any drawbacks, unless I've introduced some unintended side-effects (not been able to find any side-effects during testing).

### Applicable Issues

#130

### DIFF
--- a/settings/language-php.cson
+++ b/settings/language-php.cson
@@ -3340,7 +3340,7 @@
     'decreaseIndentPattern': '^(.*\\*/)?\\s*(\\)+)'
 '.source.php:not(.string)':
   'editor':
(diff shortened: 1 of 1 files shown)


50Wliu:
Historically there have been some issues by allowing `case` to be auto-indented, so I'm reluctant to merge this change.  See https://github.com/atom/language-javascript/issues/49.

stianlik:
I see what you mean, you have to de-indent manually after a `break;`. This is also the case in PHPStorm, which gives a very good coding experience for PHP development. I think it works well, it saves one click on the TAB-key, and it encourages developers to follow PHP conventions.

PR #330, opened 2018-05-16
Use injections for SQL strings

### LINKED ISSUE #321: Certain string literal breaks syntax parser (end of literal is not detected)
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

While working on some ancient software I noticed that certain string literal breaks syntax parser: the end of the literal is not detected correctly, which creates various side-effects in following code, like incorrect coloring or faulty bracket-matching.

The string was originally a long SQL query. I managed to shrink it down a bit, but I am unsure what combination of characters actually triggers the bug, as simple substrings of the "culprit" seem to be parsed correctly.

### Steps to Reproduce

1. Open a new editor tab
2. Switch language to PHP
3. Enter this code snippet:
    ```PHP
    <?php
    $sql = 'SELECT CONCAT(\'"\', TRIM(cr.code)) as code';
    $test = 123;
    ```

**Expected behavior:**

- String literal is correctly detected: it ends at semicolon, and both variable references (`$sql` and `$test`) are displayed the same way.
- Placing a cursor inside string literal on opening bracket highlights closing bracket inside the string literal.

**Actual behavior:**

Screenshot
- String literal does not end in correct place, but "extends" further in code (see attached screenshot). On the code sample it ends in middle of next line, but in actual code I have seen it span further, even to end of file.
- Placing a cursor on opening bracket does not highlight correct closing bracket. On the snippet it does not highlight anything, on actualy code it highights some seemingly random closing bracket further in code.

**Reproduces how often:** 100%

### Versions

OS: Linux Mint 18.2 (64bit)
Atom    : 1.25.0
Electron: 1.7.11
Chrome  : 58.0.3029.110
Node    : 7.9.0
apm  1.19.0
npm  3.10.10
node 6.9.5 x64
atom 1.25.0
python 2.7.12
git 2.7.4

50Wliu:
### Requirements

### Description of the Change

Now that language-sql has begin/end matches, injections are needed so that the PHP patterns are matched inside the begin/end ones as well.  Unfortunately, due to some annoyances with how injections work, I had to place the injections inside of the wrapper `html.cson` rather than `php.cson`, which would make much more sense.

~~I am also having trouble getting interpolation to work properly.  Seems like it's not picking up source.php correctly, which is discouraging...~~

### Alternate Designs

I would love to hear proposals for alternate designs.

### Benefits

Embedded SQL strings should no longer have the chance of ruining syntax highlighting for the rest of the file.

### Possible Drawbacks

Grammar becomes harder to parse due to the injections being in a separate file.

### Applicable Issues

Fixes #321

### DIFF
--- a/grammars/html.cson
+++ b/grammars/html.cson
@@ -48,6 +48,83 @@
         'include': '#php-tag'
       }
     ]
+  'L:source.php string.quoted.single.sql.php source.sql.embedded.php':
+    'patterns': [
+      {
+        'match': '(#)(\\\\\'|[^\'])*(?=\'|$)'
+        'name': 'comment.line.number-sign.sql'
+        'captures':
+          '1':
+            'name': 'punctuation.definition.comment.sql'
+      }
+      {
+        'match': '(--)(\\\\\'|[^\'])*(?=\'|$)'
+        'name': 'comment.line.double-dash.sql'
+        'captures':
+          '1':
+            'name': 'punctuation.definition.comment.sql'
+      }
+      {
+        'match': '\\\\[\\\\\'`"]'
+        'name': 'constant.character.escape.php'
+      }
+      {
+        # Unclosed strings must be captured to avoid them eating the remainder of the PHP script
+        # Sample case: $sql = "SELECT * FROM bar WHERE foo = \'" . $variable . "\'"'
+        'match': '"(?=((\\\\")|[^"\'])*(\'|$))'
+        'name': 'string.quoted.double.unclosed.sql'
+      }
+    ]
+  'L:source.php string.quoted.double.sql.php source.sql.embedded.php':
+    'patterns': [
+      {
+        'match': '(#)(\\\\"|[^"])*(?="|$)'
+        'name': 'comment.line.number-sign.sql'
+        'captures':
+          '1':
+            'name': 'punctuation.definition.comment.sql'
+      }
+      {
+        'match': '(--)(\\\\"|[^"])*(?="|$)'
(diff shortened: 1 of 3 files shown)


Ingramz:
@50Wliu I took a quick look at it and it seems like the culprit is that you are trying to include from a repository of a different grammar inside the injection: `source.php#interpolation_double_quoted`.

When I copied these repository items to the html grammar's repository and included those instead, it seemed to work without issues. Indirectly including `source.php#interpolation_double_quoted` from local repository within the injection did not work either.

Whether it is supposed to work in TextMate remains to be seen once I get home from work.

Overall, the idea of using injections seems suitable for at least PHP. Having to do it from the html grammar is somewhat suboptimal, but since nested injections definitely do not work in TextMate, this is probably the best chance we got if it works.

Ingramz:
My tests with TextMate show that including from other grammar's repository should work in injections too, so it is likely that something needs to be fixed in first-mate to get this to work.

PR #350, opened 2019-01-21
Add missing PHP code mixing within <style> tags

esemlabel:
### Description of the Change
CSS one of the required languages that can be mixed within PHP files along with HTML and Javascript, needed to embed specific code based on conditions.

HTML mixing `<div><?php echo $title ?></div>` and JS mixing `<script type="text/javascript"><?php echo $var ?></script>` inline or within blocks have been already implemented long time ago. But CSS support for `<style>` tags still missing.

CSS mixing is required only for php type files (php5, phtml, etc.) and only within `<style>` tags.
Mixing inside .css files, as well as .html and .js, is not allowed.

### Alternate Designs
The only way to mixing css styles without breaking syntax highlighting is to duplicate whole `<style>` tags with different content or force to use external stylesheets, which is not always applicable when using templates, MVC pattern, when avoiding render blocking, etc.

### Benefits
PHP parser not depends on IDE or any editor syntax highlighting. But this pull request adds support for correct syntax recognizing inside Atom, VS Code and many other editors that uses Atom's grammar, which needed for syntax linters, intellisense, code snippets and suggestions to work as well inside mixing content.

### Possible Drawbacks
VS Code, for which this pull request was made initially, uses modified grammar rules.
VS Code rule `L:(source.css - (meta.embedded.block.php | meta.embedded.line.php))` can be loaded from /php/ folder extension, but Atom rules may require modifications to work only within php files. This may be `source.css.embedded.html` if there is exists one.

### Applicable Issues
None.

### Example
```
<style type="text/css">
<?php if ( empty( $comments ) ) { ?>
.comments {
    display: block;
}
<?php } ?>
</style>
```

### DIFF
--- a/grammars/html.cson
+++ b/grammars/html.cson
@@ -42,7 +42,7 @@
 'foldingStartMarker': '(/\\*|\\{\\s*$|<<<HTML)'
 'foldingStopMarker': '(\\*/|^\\s*\\}|^HTML;)'
 'injections':
-  'text.html.php - (meta.embedded | meta.tag), L:((text.html.php meta.tag) - (meta.embedded.block.php | meta.embedded.line.php)), L:(source.js.embedded.html - (meta.embedded.block.php | meta.embedded.line.php))':
+  'text.html.php - (meta.embedded | meta.tag), L:((text.html.php meta.tag) - (meta.embedded.block.php | meta.embedded.line.php)), L:(source.js.embedded.html - (meta.embedded.block.php | meta.embedded.line.php)), L:(source.css - (meta.embedded.block.php | meta.embedded.line.php))':
     'patterns': [
       {
         'include': '#php-tag'
--- a/spec/html-spec.coffee
+++ b/spec/html-spec.coffee
@@ -109,6 +109,34 @@ describe 'PHP in HTML', ->
       expect(lines[7][1]).toEqual value: 'script', scopes: ['text.html.php', 'meta.tag.script.html', 'entity.name.tag.script.html']
       expect(lines[7][2]).toEqual value: '>', scopes: ['text.html.php', 'meta.tag.script.html', 'punctuation.definition.tag.html']
 
+    it 'does not tokenize PHP tag syntax within PHP syntax itself inside HTML style tags (regression)', ->
+      lines = grammar.tokenizeLines '''
+        <style>
+        <?php
+          /*
+            ?>
+            <?php
+          */
+        ?>
+        </style>
+      '''
+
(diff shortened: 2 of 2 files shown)


rsese:
Thanks @esemlabel!  Can you edit the issue body with the [pull request template](https://github.com/atom/language-php/blob/master/PULL_REQUEST_TEMPLATE.md) details?  The format and information asked for is helpful for anyone that has to review a pull request.

KapitanOczywisty:
@rsese Pull request was updated a long time ago and should be merged in next batch. I've checked this change on `atom` and `vscode` - works as expected.

_This fixes: https://github.com/microsoft/vscode/issues/33519_

rsese:
Thanks for the heads up @KapitanOczywisty, there aren't notification for edits so missed the pull request body update (sorry about that @esemlabel!).

Now that the PR body has been updated to use the template, I noticed there aren't any tests, is that something you can add @esemlabel?  If so, can you drop a comment after that's done so we get a notification?

> should be merged in next batch

Just wanted to mention that as with other pull requests, we can't promise any specific timeframe for reviewing or merging a particular PR.

esemlabel:
@rsese , what kind of test do you mean?
Currently, I'm using this patch for the last year in my development and didn't face up with any issues.

KapitanOczywisty:
@esemlabel Something like that, just for `style` tag https://github.com/atom/language-php/blob/e3281a842ffed11adbe8763b0ddcffcf308a30fb/spec/html-spec.coffee#L83-L111

esemlabel:
@KapitanOczywisty,
Unfortunately, I'm not familiar to coffee script or any other exposing tokenized lines technique.

KapitanOczywisty:
Merge me or copy updated file
https://github.com/KapitanOczywisty/language-php-style/commit/bf95a9cb470ac7531f659f144c2d40e5035d7b53

https://github.com/KapitanOczywisty/language-php-style/blob/bf95a9cb470ac7531f659f144c2d40e5035d7b53/spec/html-spec.coffee#L112-L138

esemlabel:
Done.

KapitanOczywisty:
VScode team agreed to merge this change in update script downstream. I've checked atom and there is no `source.css` scope, so change would have no effect in atom itself. VScode is overriding injections, so this change would need to be done there anyway.

This pr can be close, and sorry that I didn't check scopes before I dug out this.

rsese:
Oh ok no problem, thanks for clarifying - since you're the pull request author @esemlabel does this sound good to you?

esemlabel:
I'm good.

rsese:
Thanks, I'll go ahead and close this out :+1:

Issue #351, opened 2019-01-21
grammar begin/match


roblourens:
Should this be `match` instead of `begin`?

https://github.com/atom/language-php/blob/608077bba25b38fdb4180af659214c2603fbbb78/grammars/php.cson#L2575

Ingramz:
It can be either per specification, meaning if there is no `end`, it won't loop. But it would be less confusing if it was `match` instead.

roblourens:
Fixed in #352

PR #352, opened 2019-01-21
begin -> match

roblourens:
### Description of the Change

As discussed in https://github.com/atom/language-php/issues/351, this rule uses `begin` but it would be less confusing and more consistent with the other rules if it used `match`. This does not change the behavior of the rule.

### Alternate Designs

N/A

### Benefits

Code clarity and consistency

### Possible Drawbacks

None

### Applicable Issues

https://github.com/atom/language-php/issues/351

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -2750,7 +2750,7 @@
         'name': 'support.function.com.php'
       }
       {
-        'begin': '(?i)\\b(isset|unset|eval|empty|list)\\b'
+        'match': '(?i)\\b(isset|unset|eval|empty|list)\\b'
         'name': 'support.function.construct.php'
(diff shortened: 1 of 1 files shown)


rsese:
Thanks for opening a pull request for this @roblourens :bow:

Can you edit the issue body to use the [pull request template](https://github.com/atom/language-php/blob/master/PULL_REQUEST_TEMPLATE.md)?  I know it's a small change so brief is totally fine but the template format and details makes things much easier on the reviewer.

roblourens:
Certainly, done 👍

rsese:
Thank you! :bow:  Someone from the team will take a look as soon as they can.

PR #357, opened 2019-04-02
Fixing syntax highlighting issue for flexible HEREDOC and NOWDOC in PHP 7.3

mohamamdreza1000:
### Description of the Change

As already reported in [346](https://github.com/atom/language-php/issues/346) flexible HEREDOC and NOWDOC as introduced in PHP 7.3 will cause syntax highlighting to be failed.

### Benefits

We can have syntax highlighting to be working even if we use flexible HEREDOC and NOWDOC.

### Possible Drawbacks

The only change which I made was the regex of HEREDOC and NOWDOC endings, so I do not think any particular issue it might cause. As it is most likely how PHP mechanism will detect the end of these two blocks.

### So :
`
$example = foo(
	<<<HEREDOC
		This is just a cool test.
	HEREDOC,
	$bar
);

$rest_of_the_code = 'hello';
`
Should now have syntax highlighting be worked properly.


wecc:
This PR seems to fail because of the `does not match incorrect heredoc terminators` test in `php-spec.coffee`.

I would suggest this test to be deleted completely since that code is now valid syntax with PHP 7.3?

rsese:
Thanks for the contribution @mohamamdreza1000!  Before the team can review, the CI failure needs to be addressed, can you take a look (see @wecc's [comment](https://github.com/atom/language-php/pull/357#issuecomment-479940309))?   Additionally, can you add tests for this so it's not broken by future changes?

mohamamdreza1000:
So, I did what @wecc had recommended. It seemed two tests had to be removed.
This HEREDOC seemed very important to me because I'm developing a PHP framework using new feature of PHP.

About adding new testing ... As I'm thinking about it:
I know the only thing which might not be acceptable by PHP 7.3 is invalid indentation in HEREDOC and NOWDOC ...
But the semicolon with additional space, in next line and HEREDOC terminator containing comment line at the end of it, are all accepted.

And about adding more tests @rsese, I simply do not know what to add as a test considering what I said in prev. paragraph.

Ingramz:
These two cases behave incorrectly right now
```php
<?php

$heredoc = <<<MYHEREDOC
                hello
    asd            MYHEREDOC;

echo $heredoc;
```
Fix this by adding `^[ \\t]*` in front of the ending marker patterns.


```php
<?php

$heredoc = <<<MYHEREDOC
                hello
                MYHEREDOCs
                MYHEREDOC;

echo $heredoc;
```
Fix this by checking that the token following is not a valid heredoc marker: `(?![a-z0-9_\\x{7f}-\\x{7fffffff}])`

Example of the final pattern: `^[ \\t]*(\\2)(?![a-z0-9_\\x{7f}-\\x{7fffffff}])`

spec/php-spec.coffee
```diff
@@ -1899,21 +1899,6 @@ describe 'PHP grammar', ->
     expect(lines[2][0]).toEqual value: 'HEREDOC', scopes: ['source.php', 'string.unquoted.heredoc.php', 'keyword.operator.heredoc.php']
     expect(lines[3][0]).toEqual value: ';', scopes: ['source.php', 'punctuation.terminator.expression.php']
 
-  it 'does not match incorrect heredoc terminators', ->
```
Ingramz: Instead of removing these tests, they can be repurposed as valid examples of a heredoc.

mohamamdreza1000: Thank you I'll try to run tests in order to make what you said earlier work properly.

Ingramz:
Looking good! I'd suggest adding one more test that is significantly different from a standard heredoc, like the example from the first post in this pull request, where the heredoc is used inside a function call.

```php
foo(<<<HEREDOC
  This is just a cool test.
  HEREDOC,
  $bar
);
```

This way there is an example of flexible heredoc in actual use, plus it tests the indentation of the ending marker, something that was not previously allowed.

Once that is done, I think we can ask this to get reviewed by someone from the team and then merged.

spec/php-spec.coffee
```diff
@@ -1863,6 +1863,28 @@ describe 'PHP grammar', ->
     expect(tokens[4]).toEqual value: '\'', scopes: ['source.php', 'string.quoted.single.php', 'punctuation.definition.string.end.php']
     expect(tokens[5]).toEqual value: ';', scopes: ['source.php', 'punctuation.terminator.expression.php']
 
+  it 'should tokenize heredoc in a function call correctly', ->
+    lines = grammar.tokenizeLines  '''
+    foo(
+    <<<HEREDOC
```
Ingramz: Indent the lines inside foo() by 2 or 4 spaces, right now it looks like this:

```php
foo(
<<<HEREDOC
This is just a cool test.
HEREDOC,
$bar
);
```

but we want

```php
foo(
  <<<HEREDOC
  This is just a cool test.
  HEREDOC,
  $bar
);
```

mohamamdreza1000: @Ingramz Looks like failure ... But without indention checks were good.

Ingramz: You'll need to update expect()... lines too, because now there is one extra token in front of every line.

Ingramz:
Add missing lines[x][0] entries as well, this should help you understand why also lines[2][0] still is failing.

mohamamdreza1000:
> Add missing lines[x][0] entries as well, this should help you understand why also lines[2][0] still is failing.

I believe string.unquoted.heredoc.php knows spaces as part of string this might be the reasons of failures ...

spec/php-spec.coffee
```diff
@@ -1863,6 +1863,31 @@ describe 'PHP grammar', ->
...
+      <<<HEREDOC
+      This is just a cool test.
+      HEREDOC,
+      $bar
+    );
+    '''
+
+    expect(lines[0][0]).toEqual value: 'foo', scopes: ['source.php', 'meta.function-call.php', 'entity.name.function.php']
+    expect(lines[0][1]).toEqual value: '(', scopes: ['source.php', 'meta.function-call.php', 'punctuation.definition.arguments.begin.bracket.round.php']
+    expect(lines[1][0]).toEqual value: ' ', scopes: ['source.php', 'meta.function-call.php']
```
Ingramz: Here needs to be two spaces instead of one.

spec/php-spec.coffee
```diff
@@ -1863,6 +1863,31 @@ describe 'PHP grammar', ->
...
+    expect(lines[1][1]).toEqual value: '<<<', scopes: ['source.php', 'meta.function-call.php', 'string.unquoted.heredoc.php', 'punctuation.definition.string.php']
+    expect(lines[1][2]).toEqual value: 'HEREDOC', scopes: ['source.php', 'meta.function-call.php', 'string.unquoted.heredoc.php', 'keyword.operator.heredoc.php']
+    expect(lines[2][0]).toEqual value: 'This is just a cool test.', scopes: ['source.php', 'meta.function-call.php', 'string.unquoted.heredoc.php']
+    expect(lines[3][0]).toEqual value: ' ', scopes: ['source.php', 'meta.function-call.php']
```
Ingramz: Here needs to be two spaces instead of one.

spec/php-spec.coffee
```diff
@@ -1863,6 +1863,31 @@ describe 'PHP grammar', ->
...
+    expect(lines[3][1]).toEqual value: 'HEREDOC', scopes: ['source.php', 'meta.function-call.php', 'string.unquoted.heredoc.php', 'keyword.operator.heredoc.php']
+    expect(lines[3][2]).toEqual value: ',', scopes: ['source.php', 'meta.function-call.php', 'punctuation.separator.delimiter.php']
+    expect(lines[4][0]).toEqual value: ' ', scopes: ['source.php', 'meta.function-call.php']
```
Ingramz: Here needs to be two spaces instead of one.

spec/php-spec.coffee
```diff
@@ -1863,6 +1863,31 @@ describe 'PHP grammar', ->
...
+      $bar
+    );
+    '''
+
+    expect(lines[0][0]).toEqual value: 'foo', scopes: ['source.php', 'meta.function-call.php', 'entity.name.function.php']
+    expect(lines[0][1]).toEqual value: '(', scopes: ['source.php', 'meta.function-call.php', 'punctuation.definition.arguments.begin.bracket.round.php']
+    expect(lines[1][0]).toEqual value: ' ', scopes: ['source.php', 'meta.function-call.php']
+    expect(lines[1][1]).toEqual value: '<<<', scopes: ['source.php', 'meta.function-call.php', 'string.unquoted.heredoc.php', 'punctuation.definition.string.php']
+    expect(lines[1][2]).toEqual value: 'HEREDOC', scopes: ['source.php', 'meta.function-call.php', 'string.unquoted.heredoc.php', 'keyword.operator.heredoc.php']
+    expect(lines[2][0]).toEqual value: 'This is just a cool test.', scopes: ['source.php', 'meta.function-call.php', 'string.unquoted.heredoc.php']
```
Ingramz: Add two spaces in front of the word "This"

spec/php-spec.coffee
```diff
@@ -1875,15 +1875,12 @@ describe 'PHP grammar', ->
 
     expect(lines[0][0]).toEqual value: 'foo', scopes: ['source.php', 'meta.function-call.php', 'entity.name.function.php']
     expect(lines[0][1]).toEqual value: '(', scopes: ['source.php', 'meta.function-call.php', 'punctuation.definition.arguments.begin.bracket.round.php']
-    expect(lines[1][0]).toEqual value: ' ', scopes: ['source.php', 'meta.function-call.php']
     expect(lines[1][1]).toEqual value: '<<<', scopes: ['source.php', 'meta.function-call.php', 'string.unquoted.heredoc.php', 'punctuation.definition.string.php']
     expect(lines[1][2]).toEqual value: 'HEREDOC', scopes: ['source.php', 'meta.function-call.php', 'string.unquoted.heredoc.php', 'keyword.operator.heredoc.php']
-    expect(lines[2][0]).toEqual value: ' ', scopes: ['source.php', 'meta.function-call.php', 'string.unquoted.heredoc.php']
+    expect(lines[2][0]).toEqual value: '  ', scopes: ['source.php', 'meta.function-call.php', 'string.unquoted.heredoc.php']
     expect(lines[2][1]).toEqual value: 'This is just a cool test.', scopes: ['source.php', 'meta.function-call.php', 'string.unquoted.heredoc.php']
```
Ingramz: This is just one token with value `'  This is just a cool test.'`

spec/php-spec.coffee
```diff
@@ -1863,6 +1863,28 @@ describe 'PHP grammar', ->
...
+      <<<HEREDOC
+      This is just a cool test.
+      HEREDOC,
+      $bar
+    );
+    '''
+
+    expect(lines[0][0]).toEqual value: 'foo', scopes: ['source.php', 'meta.function-call.php', 'entity.name.function.php']
+    expect(lines[0][1]).toEqual value: '(', scopes: ['source.php', 'meta.function-call.php', 'punctuation.definition.arguments.begin.bracket.round.php']
+    expect(lines[1][1]).toEqual value: '<<<', scopes: ['source.php', 'meta.function-call.php', 'string.unquoted.heredoc.php', 'punctuation.definition.string.php']
```
Ingramz: lines[1][0] is missing

mohamamdreza1000: > lines[1][0] is missing

I just removed those intentions because I thought it was unnecessary as earlier examples omitted indention as well. if u think it is better to add them so I will ...

Ingramz: That's fine then, my apologies. The offset should be enough to indicate if something's wrong in the future.

mohamamdreza1000: And btw,
Since I'm new to whole GitHub, Do I need to force push and make a cleaner commit or just leave things as they are right now ?

Ingramz: You can do a force push if you want to. But users who have the ability to merge pull requests, can choose to squash noisy pull requests themselves. See https://help.github.com/en/articles/merging-a-pull-request#merging-a-pull-request-on-github

mohamamdreza1000: I guess, I learned enough for now ...
I've been a programmer since 14 and I was afraid of git till recent year.
Thank you @Ingramz

Ingramz:
@rsese this pull request should now be ready for final review/merge.

mohamamdreza1000:
My mistakes of making to many commits and not knowing how to merge them ... Thanks to guides I finally made it you may now reopen the pull.
Big thanks to @Ingramz for helping me.

Ingramz:
Something went wrong when force pushing, checkout a clean branch from this repo and apply your commits on top of that, then force push it as the branch that you use for this pull request.

PR #359, opened 2019-04-06
Revert SQL-related changes from last year

### LINKED ISSUE #333: sql statement breaks syntax highlighting
### Prerequisites

* [X ] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

There seems to be an issue parsing a line of SQL code that results in the rest of the file being treated as a "string".

### Steps to Reproduce

1. Open this file in atom: https://github.com/nosilver4u/ewww-image-optimizer/blob/4944e9677f2d529e429ca6d87ebcd1664997e32c/common.php#L5193
2. scroll to line 5193
3. note the syntax highlighting goes wonky about halfway through that line.

### Versions

I'm running 1.27.2 on MacOS. The issue also occurred before I updated yesterday, and I believe I was running 1.25. On my linux desktop, running 1.23, the issue does not happen.

Ingramz:
### Description of the Change

This pull request reverts pull requests #330 and #332 due to causing more harm than good.

While in ideal conditions these pull requests do an amazing job at highlighting in SQL in PHP strings, it can break very easily by having invalid SQL in PHP strings. Do note that PHP has never required strings to contain valid SQL and never will, so it is incorrect of us to assume that.

### Alternate Designs

Attempt fixing the current implementation: neither @50Wliu or I have the time to deal with this and it is fairly evident from pull requests that nobody else dares to touch it either. Since originally the intent was to fix any issues as they appear, this definitely does not apply today, so reverting is the next best thing to do.

### Benefits

SQL may only be parsed inside PHP strings, leaving rest of the code to PHP grammar to parse.

### Possible Drawbacks

In some situations we are going a step backwards due to some lost ability to highlight SQL more accurately. If anyone desires, these can be re-implemented afterwards, however they do require more thorough review before getting merged.

### Applicable Issues

Fixes #333


Ingramz:
Here's another idea, add the currently failing cases as tests, so they wouldn't break again.

Ingramz:
https://github.com/atom/language-sql/pull/70 is also part of the culprit. Python suffers from a similar issue, where a set of parentheses remains unclosed within same string:

```php
<?php
echo "hello world";
$sql = "SELECT * from (";
echo "hello world";
```

image


```python
print "hello world"
sql = "SELECT * from ("
print "hello world"
```

image

lee-dohm:
@50Wliu do you have any thoughts on this?

50Wliu:
> Here's another idea, add the currently failing cases as tests, so they wouldn't break again.

This should be done.

> [atom/language-sql#70](https://github.com/atom/language-sql/pull/70) is also part of the culprit.

What would be the impact of reverting atom/language-sql#70?

Ingramz:
> What would be the impact of reverting [atom/language-sql#70](https://github.com/atom/language-sql/pull/70)?

I'm not entirely sure, probably not much as it is the latest grammar related change on that repository. It will definitely reintroduce the issue it was supposed to fix, but otherwise looks harmless.

50Wliu:
I think atom/language-sql#70 was the main cause of this issue due to its introduction of begin/end captures. Maybe we can try reverting the punctuation-related changes of that one first?

Ingramz:
Yes, probably. I agree that we should try that first.

I'll take a closer look at the left-injections of this pull request later and then analyze whether there's anything really wrong with them. But it is obvious now that language-sql change had a way bigger impact in terms of breaking highlihting in this case.

sadick254 (approved):

KapitanOczywisty:
@sadick254  I'm not sure if this should be merged, It's very old and may break sql even more...

KapitanOczywisty:
After some more testing seems that some issues were indeed "fixed" by this revert (besides broken stuff), but what is funny vscode have also fixed issues and at the same time I haven't find any degradations yet. With that I will not bother to fix any SQL issues in atom anymore.

atom_before
atom_after

And vscode changes

Code_before
Code_after

Issue #225, opened 2017-06-25
Keyword after instanceof should be given a unique scope


jens1o:
### Prerequisites

* [x] I'm using vscode

### Description

![image](https://user-images.githubusercontent.com/11234139/27514948-e5c92e24-5998-11e7-93f0-ee48ff27f5c5.png)

### Steps to Reproduce

```php
<?php
// Rethrow anything that is not a RuntimeException
if (!$e instanceof RuntimeException) {
    throw $e;
}

50Wliu:
Can't reproduce, and it does seem to work in your picture unless I'm misunderstanding something.  It's being tokenized as `keyword.operator.type.php` for me.

jens1o:
For me it gets tokenized for an `constant.other.php`...
![image](https://user-images.githubusercontent.com/11234139/28002379-ef4f6cfe-6534-11e7-9ecd-85b31b681d90.png)
@aeschli Please fix the vscode's grammar parser... :/

50Wliu:
Oh, do you mean `RuntimeException` is being tokenized as a constant, or that `instanceof` is?  Or perhaps, what exactly do you mean by "instanceof does not work"?

jens1o:
Yes, that's what I mean, it shouldn't be a constant. Sorry, I'm in a hurry.

50Wliu:
No problem.  What do you think it should be tokenized as then?

jens1o:
`keyword.operator.type.php`... But it's working fine for you. Maybe the vscode grammar parser has a bug?

aidantwoods:
I concur that `RuntimeException` should be a class, and not a const.

More detail: https://github.com/Microsoft/vscode/issues/30500#issuecomment-318504496

mattacosta:
I don't think that it should be under the `keyword` group. It makes more sense to keep it similar to a type that comes after the `new` keyword. So probably `support.class.php` or `support.class.instanceof.php` if necessary.

Also, this issue should be marked as a bug, since PHP never parses an `instanceof` type as a constant.

PR #384, opened 2020-03-05
Parse class with capitol letter after instanceof correctly

roblourens:
### Description of the Change

It looks like the original author of this section did intend to parse the thing after `instanceof` correctly but only included `a-z` not `A-Za-z`. You can see that this does work as expected with class names starting with a lowercase letter.

### Benefits

Show correct coloring for the class name

### Possible Drawbacks

I think this is fairly straightforward

### Applicable Issues

atom/language-php#225
microsoft/vscode#88441

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -728,11 +728,11 @@
     'name': 'keyword.operator.bitwise.php'
   }
   {
-    'begin': '(?i)\\b(instanceof)\\s+(?=[\\\\$a-zA-Z_])'
+    'begin': '(?i)\\b(instanceof)\\s+(?=[\\\\$a-z_])'
     'beginCaptures':
       '1':
         'name': 'keyword.operator.type.php'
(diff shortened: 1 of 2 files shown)


grammars/php.cson
```diff
@@ -728,11 +728,11 @@
     'name': 'keyword.operator.bitwise.php'
   }
   {
-    'begin': '(?i)\\b(instanceof)\\s+(?=[\\\\$a-z_])'
+    'begin': '(?i)\\b(instanceof)\\s+(?=[\\\\$a-zA-Z_])'
```
Ingramz: Modifier in front is already marks this expression case insensitive, so this change should not be needed.

roblourens: It does fix the issue though, hm...

roblourens: Oh, only the second line is needed

Ingramz: Yeah, I realized that it could be read like that, sorry about that.

darangi:
Thanks for the contribution!

Issue #395, opened 2020-07-23
PHP 8.0


KapitanOczywisty:
Bellow you will find list of new features added in PHP 8.0, which are affecting **syntax**.

I'll be creating PR's, but feel free to jump in and add features. If you want to help, but you have problems with TM grammar or test specs or anything, just let me know in this issue.

🏃 - Work in progress
🐌 - Planned (some work started)
✋ - On hold (possible changes)
⏳ - PR waiting for review & merge
✔️ - PR merged

- PRs
  - #396 ✔️ 
    - [x] https://wiki.php.net/rfc/non-capturing_catches
  - #398 ✔️ 
    - [x] https://wiki.php.net/rfc/mixed_type_v2 
    - [x] https://wiki.php.net/rfc/trailing_comma_in_parameter_list 
    - [x] https://wiki.php.net/rfc/static_return_type 
    - [x] https://wiki.php.net/rfc/union_types_v2 
    - [x] https://wiki.php.net/rfc/constructor_promotion 
    - [x] https://wiki.php.net/rfc/trailing_comma_in_closure_use_list
  - #399 ✔️ 
    - [x] https://wiki.php.net/rfc/named_params
  - #404 ✔️ 
    - [x] https://wiki.php.net/rfc/nullsafe_operator
  - #405 ✔️ 
    - [x] Attributes (final syntax `#[Attribute]`)
      - https://wiki.php.net/rfc/attributes_v2
      - https://wiki.php.net/rfc/attribute_amendments
      - https://wiki.php.net/rfc/shorter_attribute_syntax
      - https://wiki.php.net/rfc/shorter_attribute_syntax_change
  - #407 ✔️ 
    - [x] https://wiki.php.net/rfc/match_expression_v2

- Other
  - [ ] Update names of constants, functions, classes 🐌 

Migration guide for PHP 8.0 is available here: https://github.com/php/php-src/blob/PHP-8.0/UPGRADING

KapitanOczywisty:
@rsese Hello sir, can you mention internally [my PRs in this repository](https://github.com/atom/language-php/pulls/KapitanOczywisty)? There is 5 of them atm and would be nice to merge them before collisions starts to be an issue. PHP 8.0 is not released yet but it's in beta and in feature freeze, so there will be no changes to syntax. Everything is backward compatible anyway.

KapitanOczywisty:
[...]

Issue #390, opened 2020-04-20
Coloring/Highlighting Issues for Exceptions and Constants in PHP


alexr00:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: <https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode>
    * Followed all applicable steps in the debugging guide: <https://flight-manual.atom.io/hacking-atom/sections/debugging/>
    * Checked the FAQs on the message board for common solutions: <https://discuss.atom.io/c/faq>
    * Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom>
    * Checked that there is not already an Atom package that provides the described functionality: <https://atom.io/packages>

### Steps to Reproduce

1. Multiple Exception types in a catch block separated by a single pipe '|' highlights only the class name of the last class (see item1.png):

example: catch (\Namespace\Exception1 | \Namespace\Exception2 | \Namespace\Exception3 $e) {...}

- only Exception3 would have correct class coloring/highlighting

![item1](https://user-images.githubusercontent.com/25559097/79602440-b737c280-80a7-11ea-8364-3a2f96afd963.png)

2. When using a namespaced constant as a default value for a parameter in a function definition is colored red incorrectly (see item2.png).

example: public static function redirect(string $locationUri, array $responseCode = \Http\Response\Code::SEE_OTHER, array $additionalRedirectHeaders = []): void

- \Http\Response\Code::SEE_OTHER is colored red even though there is no error.  changing this to a [] colors correctly, but the other does not

![item2](https://user-images.githubusercontent.com/25559097/79602465-c159c100-80a7-11ea-8441-f988301e6f6b.png)

3. When using an Uppercase constant name that is the same as a PHP primitive (in my example: STRING vs string, the coloring is incorrect.  It is colored as a type instead of the constant name (see item3.png)

example: public const STRING = 2;

- String is colored as if it were a data type primitive instead of the color of constant declared names.

![item3](https://user-images.githubusercontent.com/25559097/79602482-c6b70b80-80a7-11ea-8be7-83e3eba0d73b.png)

**Expected behavior:**

**Actual behavior:**

**Reproduces how often:**

### Versions

Version 0.44.4 from commit 882f6c0e19f0ebf9dafa443bf4c3fc5626f76aed

### Additional Information

- VSCode Version: 1.44.1
[...]

PR #396, opened 2020-08-11
Non-capturing catches and multiple namespaced exceptions

KapitanOczywisty:
This is part of #395 

RFC: https://wiki.php.net/rfc/non-capturing_catches

There is no backward incompatible changes, safe to merge

Also fixes partially #390 (1st point)


KapitanOczywisty:
I've added fix for one of the issues in #390 , since it's the same part of code.

PR #405, opened 2020-11-27
Attributes

KapitanOczywisty:
This is implementation of Attributes added in PHP 8.0. However documentation for this feature is rather bad, so I'd advice you to use sth like https://3v4l.org/ to determinate how this feature really works.

### Implemented:
- argument-less attribute `#[MyAttribute]`
- attribute with arguments `#[MyAttribute(true)]`
- attribute with named arguments `#[MyAttribute(name:'Bob')]`
- namespaced attributes `#[\Ye\Old\Attribute]`
- multiple attributes `#[A,B,C]`
- multilined attributes 
```
#[
A,
B(true)
]
```
- anonymous class attributes (single-line only) `new #[MyAttribute] class {}`, also it's valid to write `new#[MyAttribute]class{}`
- `Attribute` as class name, for use as argument `#[Attribute(Attribute::TARGET_CLASS)]`
- test spec

### Scopes
- `meta.attribute.php` for whole `#[....]`
- `support.attribute.php` for attribute name, w/ namespace scopes like in classes
- `support.attribute.builtin.php` for `Attribute` and future builtins
- arguments are the same as in function call

### Not implemented
- multilined attributes for anonymous classes - I'm not sure if I should implement this, since it would need some trickery to add with TM grammar (feedback is welcome)

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -1032,10 +1032,7 @@
         ]
       }
       {
-        'match': '[a-z0-9_\\x{7f}-\\x{7fffffff}\\\\]+'
-        'patterns': {
-          'include': '#attribute-name'
-        }
+        'include': '#attribute-name'
       }
     ]
   'class-builtin':
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -9,6 +9,9 @@
 'scopeName': 'source.php'
 'patterns': [
   {
+    'include': '#attribute'
+  }
+  {
     'include': '#comments'
   }
   {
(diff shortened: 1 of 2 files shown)


grammars/php.cson
```diff
@@ -206,7 +206,7 @@
   {
     'begin':  '''(?ix)
         (?:
-          (?:^|(?<=}))\\s*(?:(abstract|final)\\s+)?(class)\\s+([a-z_\\x{7f}-\\x{7fffffff}][a-z0-9_\\x{7f}-\\x{7fffffff}]*)
+          \\b(?:(abstract|final)\\s+)?(class)\\s+([a-z_\\x{7f}-\\x{7fffffff}][a-z0-9_\\x{7f}-\\x{7fffffff}]*)
```
KapitanOczywisty, line 209: I've removed complicated look-behind. As far I'm concerned, there is no case where this could backfire, but I could be wrong.

KapitanOczywisty:
@sadick254 ready for review 😃

sadick254 (approved):
Looks great 👍

PR #406, opened 2020-11-27
Update .travis.yml

KapitanOczywisty:
Fix for recent build failures

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ git:
 
 sudo: false
 
-dist: trusty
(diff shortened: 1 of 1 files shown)


KapitanOczywisty:
@sadick254 can you merge this?

sadick254:
@KapitanOczywisty we could update `gcc/g++` on `xenial`. The queuing is something external and could change, we should not base our solutions on external factors we have no control over. I am happy with this change. Thank you @KapitanOczywisty .

Issue #92, opened 2015-07-30
Double-click highlights whole of variable


Sogl:
Hi!

If I double-click on variable in PHP, than it highlights the whole of the variable.
See also:
https://discuss.atom.io/t/double-click-highlights-whole-of-variable-in-php/18625
https://github.com/atom/language-php/issues/43

As I can see, this problem is fixed (https://github.com/atom/language-php/pull/75), but for me it still present in last atom version.

Also tried to put `/\\()"\':,.;<>~!@#%^&*|+=[]{}`?-` in non word characters, but nothing changed =(

Env:
- Windows 8.1 pro
- Atom 1.0.3
- language-php 0.27.0

kevinsawicki:
> Also tried to put /\\()"\':,.;<>~!@#%^&*|+=[]{}`?- in non word characters, but nothing changed =(

Thanks for reporting this, this was definitely a bug which will be fixed in the next Atom release.

Until the next release here is how you can workaround it, run Atom > Open Your Config and in the opened file add the following:

```coffee
".source.php":
  editor:
    nonWordCharacters: "/\\\\()\"\\':,.;<>~!@#%^&*|+=[]{}`?-$"
```

This should no longer select the `$` character when double clicking

Sogl:
>This should no longer select the $ character when double clicking

Yep, it works. Thanks, @kevinsawicki !

abejfehr:
@kevinsawicki This is still not present in the newest version of `language-php`. Was this forgotten?

Sogl:
@kevinsawicki @abejfehr Confirmed, doesn't work in last Atom version.

villermen:
Any progress on this? I am still unable to select only the name of a variable. When double-clicking on `asdf` in `$asdf`, the entire string is selected even though the dollar character `$` is considered a non-word character.

Ben3eeE:
I have worked around this bug by adding `$` to the `nonWordCharacters` setting in the `language-php` package. It seems like any setting different than the default for the `nonWordCharacters` setting fixes the problem.

![image](https://cloud.githubusercontent.com/assets/1058982/20709718/3ee57fcc-b637-11e6-9125-8236c6b8b181.png)

50Wliu:
So interestingly...this behavior was changed because some people didn't like it: #43.  Can't satisfy everyone I guess 😞.

forsvunnet:
[...]

PR #412, opened 2021-01-24
Adding '$' character for highlight-selected

JesseSteele:
### Requirements

I have tested this by making a git clone into '/usr/lib/atom/node_modules' and it behaved as intended, with no bugs or warnings.

### Description of the Change

Changes the default value of of "Non Word Characters" in language-php to match the solution in the thread here:

https://discuss.atom.io/t/double-click-highlights-whole-of-variable-in-php/18625

The thread above suggests (from 2015):

```
$/()"’:,.;<>~!@#%^&*|+=[]{}`?-
```

However, adding a `$` character to the current settings would result in the requested change here, being:

```
$/\()"':,.;<>~!@#%^&*|+=[]{}`?-
```

...This is the new value used in this pull request (2021).

### Alternate Designs

NA: There were no alternatives found because this worked on the first try. Since it was also least invasive, trials went no further.

### Benefits

Double-clicking on a PHP variable is not picked up by `highlight-selected` or its `minimap-highlight-selected` variant. This resolves that issue so double-clicking on a PHP variable not only highlights the name of the single variable, but `highlight-selected` picks up other occurrences.

### Possible Drawbacks

The main drawback would be ***if*** the problem lies within the `highlight-selected` package on a deeper coding level.

However, philosophically, this should probably be the default value anyway, as per the Atom discussion. If there is a deeper problem with the `highlight-selected` package, then this bug here helped identify that problem, but this default value should probably be changed anyway.

I'm not an expert in the full code; just doing the leg work rather than adding to other people's to-do lists. I suppose a better way might exist. If so, by all means. But, the problem wasn't solved and this solves it by the official Atom discussion. Cheers!

### Applicable Issues

This does relate to the `higlight-selected` package.

For discussion: Without this fix, the same text still highlighted on double-click. It was `highlight-selected` that failed to pick it up. So, it seems this setting only affected another package (`highlight-selected`), not the native select capability itself. That suggests that other packages may depend on this setting more than native Atom packages do. Not knowing the full code, I don't know, but it seems strange. Still, this fixed the problem by the official discussion I read.

### DIFF
--- a/settings/language-php.cson
+++ b/settings/language-php.cson
@@ -3344,4 +3344,4 @@
     'decreaseIndentPattern': '^(.*\\*/)?\\s*(}|(\\)+([;,]|\\s*{))|(\\]\\)*([;,]|$))|else:|((end(if|for(each)?|while|switch));))'
 '.text.html.php':
   'editor':
-    'nonWordCharacters': '/\\()"\':,.;<>~!@#%^&*|+=[]{}`?-'
+    'nonWordCharacters': '$/\\()"\':,.;<>~!@#%^&*|+=[]{}`?-'


KapitanOczywisty:
Double-click should highlight `variables` as a whole - with `$`. It's currently bugged, but I hope will be fixed https://github.com/atom/language-php/issues/92#issuecomment-765348037

sadick254 (approved):

keevan:
So it seems like this issue has been flip flopping for a while now. 
https://github.com/atom/language-php/issues/92#issuecomment-307540356

I personally use vim (vim-mode-plus) and find this incredibly frustrating whenever the language gets automatically updated. Where before I would `yw` to copy the `$variable`, I now have to do `ye` or some other variation. Bit worse when I'm in the middle of the word (can't just do `yiw` like I used to. I say, stop changing it, and let people who prefer one or the other configure it. If it becomes a big enough deal, add a config checkbox for it specifically. I just hope that my non-default **without** the `$` will stay even if it becomes the default again, and inevitably flips again...
```
/\()"':,.;<>~!@#%^&*|+=[]{}`?-
```

KapitanOczywisty:
@keevan VScode has also vim extension and works way better than atom, just saying...

keevan:
> @keevan VScode has also vim extension and works way better than atom, just saying...

@KapitanOczywisty I respect your opinion. Do you know if VSCode's vim extension has something similar to VMP's occurrence selector (https://github.com/t9md/atom-vim-mode-plus/wiki/OccurrenceModifier#normal-operation-vs-occurrence-modified-operation), being able to visual mark words (which does a select on the current smart-word, and you can use `caf` to change the occurence of it in a function after visually confirming it's what you wanted to change and/or jump through them), as well as atom-narrow (https://github.com/t9md/atom-narrow). I use these 2 extensively so without them I'm finding it hard to switch to anything else.

Related discussions:
https://github.com/t9md/atom-vim-mode-plus/issues/1132
https://github.com/t9md/atom-vim-mode-plus/issues/1063

That said, getting off topic from this issue. So I'll leave it at that.

PR #420, opened 2021-03-30
Fix: require(), require_once(), include(), include_once() wrong prop

andregalastri:
From: microsoft/vscode#120071
## Description of the Change
The require/include commands have two ways to use:
Without parentheses:
```php
<?php
require 'file.ext';
require_once 'file.ext';
```

And with parentheses, like this:
```php
require('file.ext');
require_once('file.ext');
```

However, the syntax highlight were only detecting the right scope of the command that doesn't use parentheses:

<table><tr><td>

</td><td>

</td></tr></table>

## Fix
This pull request resolves it by adding the **`(`** char after the **`require`**/**`require_once`** and **`include`**/**`include_once`** as a valid match to the  **`keyword.control.import.include.php`** scope.

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -351,7 +351,7 @@
         'name': 'keyword.control.${1:/downcase}.php'
   }
   {
-    'begin': '(?i)\\b((?:require|include)(?:_once)?)[\\s\\(]+'
+    'begin': '(?i)\\b((?:require|include)(?:_once)?)(?=\\s|\\()'
     'beginCaptures':
       '1':
         'name': 'keyword.control.import.include.php'
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -351,7 +351,7 @@
         'name': 'keyword.control.${1:/downcase}.php'
   }
   {
-    'begin': '(?i)\\b((?:require|include)(?:_once)?)(?=\\s|\\()'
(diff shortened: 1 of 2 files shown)


grammars/php.cson
```diff
@@ -351,7 +351,7 @@
         'name': 'keyword.control.${1:/downcase}.php'
   }
   {
-    'begin': '(?i)\\b((?:require|include)(?:_once)?)\\s+'
+    'begin': '(?i)\\b((?:require|include)(?:_once)?)[\\s\\(]+'
```
KapitanOczywisty, line 354: 1. we want to check only for single parenthesis
2. don't consume parenthesis if you're not adding token for it
for the reasons above I'd suggest lookahead:
```suggestion
    'begin': '(?i)\\b((?:require|include)(?:_once)?)(?=\\s|\\()'
```
3. You need to add some testing to prevent regression

KapitanOczywisty (changes requested):

andregalastri:
Right, sorry, this is the first time I'm contributing for a project.

I made some changes, can you take a look please? Thanks!

KapitanOczywisty:
I'm suspecting that you're not too familiar with TM syntax and atom testing. I can make pull request from mentioned commit https://github.com/KapitanOczywisty/language-php/commit/76da9e99ed5633979ee756efb83863993a1af7c5 in place of this one. Right now it's just easier for me to write fix, check it and commit, rather than guide you how to do it properly yourself.

PR #421, opened 2021-03-31
Allow include with parentheses

KapitanOczywisty:
Based on #420

PR will fix tokenization of `include`/`require` (and `_once` variants) when used with parentheses:
```php
include('foo.php');
```
Tests included


sadick254 (approved):
Thanks for adding the specs for this.

PR #424, opened 2021-04-18
More consistent tokens for types

### LINKED ISSUE #422: Inconsistent highlighting of "void"
Reported in https://github.com/microsoft/vscode/issues/120954

Is this as expected? I have no idea.

Original steps:

## Steps to Reproduce:

1. Create a new .php file
2. Paste the following code
  ```php
<?php
class MyClass
{
    public static function methodA() : void
    {
    }

    public static function methodB() : string
    {
        return "I'm a string";
    }
}
  ```
3. Check the editor tokens and scopes
image

You will see that the void type declaration is hightlighted as a support.class. I understand that it is a keyword, like `string`, `int` and others, not a class.

---

Also noticed that it's inconsistent here:

```php
<?php

/**
 * @param void $x
 */
function a($x): void {
	return '';
}
```

image

KapitanOczywisty:
I've replaced `storage.type.php` with `keyword.other.type.php` for builtin types for consistency. Also added `void` as return type. Fixes: #422

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -2320,8 +2320,8 @@
       }
       {
         # false is allowed in unions
-        'match': '(?i)\\b(null|int|float|bool|string|array|object|callable|iterable|false|mixed)\\b',
-        'name': 'storage.type.php'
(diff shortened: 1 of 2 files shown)


sadick254 (approved):
LGTM

Issue #418, opened 2021-03-08
Comment in sql string breaks highlighting


roblourens:
From https://github.com/microsoft/vscode/issues/118450

Write following Code:
```php
<?php

/*** GET ***/
$System->Router('DELETE /push/*', function ($self, $variables) {
    ...
});
$System->Router('GET /push', function ($self, $variables) {
    ...
});
```
---

Now you can see, that everything after /push/* will be handled as a comment in highlighting. 

Similar to https://github.com/atom/language-php/issues/187

KapitanOczywisty:
This shouldn't be treated as SQL, we might add some filtering to check the next word after `DELETE` to remove some false-positives. Second thing is this endless battle with embedded syntaxes, where nobody wants to touch TextMate syntax - this should be improved a long time ago..

PR #427, opened 2021-04-26
Update php.cson to detect SQL CTEs

NSExceptional:
### Description of the Change

Update `php.cson` to detect SQL CTEs by making it detect `"WITH ... "` (both single and double quoted) as an SQL string.

SQL "common table expressions" (CTE) begin with the `WITH` keyword. More info:
https://docs.microsoft.com/en-us/sql/t-sql/queries/with-common-table-expression-transact-sql?view=sql-server-ver15#examples

### Benefits

CTEs are syntax highlighted. Currently they are not, but they are still valid SQL.

### Possible Drawbacks

None

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -2697,7 +2697,7 @@
     'match': '\\\\(?:\\\\(?:\\\\[\\\\\']?|[^\'])|.)'
     'name': 'constant.character.escape.php'
   'sql-string-double-quoted':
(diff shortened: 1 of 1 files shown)


KapitanOczywisty (approved):
There is slight possibility for false-positives, but shouldn't be too serious.

https://grep.app/search?q=%5B%27%22%5DWITH%5Cs&regexp=true&case=true&filter[lang][0]=PHP

NSExceptional:
Yeah, stuff like #418 is generally inevitable without complicating the pattern. It's a tradeoff most of us are okay with I think

PR #436, opened 2021-07-22
Pure Intersection types

KapitanOczywisty:
PHP 8.1 https://wiki.php.net/rfc/pure-intersection-types
Also added intersection types in PHPDoc and some test for union types.


sadick254 (approved):
Thank you for always adding specs on your PR's. They make reviewing PR's a little easier.

PR #438, opened 2021-08-02
Update draft tree-sitter grammar in #303

claytonrcarter:
### Requirements

* ✅ Filling out the template is required.
* ✅ All new code requires tests

### Description of the Change

⚠️ This PR is targeted at the `mb-tree-sitter-parser` branch for #303 ⚠️ 

1. updates #303 to `master`, resolving conflicts
2. updates `tree-sitter-php` to v0.16.2
3. ~~ports~~will port *most* existing specs from the current TextMate grammar to run against the new tree-sitter grammar
4. updates the changes started in #303 to "finish" the tree-sitter PHP grammar
5. changes #303 to use `source.php` instead of `text.html.php`. This matches the TM grammar and fixes things like snippets.

I say that this "finishes" the tree-sitter grammar because there are still many current lanuage features that aren't available in `tree-sitter-php` 0.16.2. Updating to 0.19 (which is itself currently blocked by https://github.com/atom/atom/pull/22130) won't even help w/ these, as many of them have only been added to TS-php recently. Until this is resolved, the following language features are being parsed as errors:
- typed properties (https://github.com/tree-sitter/tree-sitter-php/pull/56)
- array spread (https://github.com/tree-sitter/tree-sitter-php/pull/64)
- arrow functions (https://github.com/tree-sitter/tree-sitter-php/pull/65)
- nullable types, catching multiple exception types, union types, `??=`, and several more

### Current Status

This is still very much ⚠️  draft ⚠️ , but appears to be usable enough for day to day work (at least for me). There are still lots of tests to port, but I expect most of these to work more or less as is. Many of TM specs may not even be needed b/c they're testing fiddly formatting issues that the `tree-sitter-php` is already handling.

**TODO**
- [ ] Finish porting tests
    - [ ] classes
        - [x] properties
        - [ ] methods
        - [ ] use statements
        - [ ] anonymous classes
    - [ ] functions
    - [ ] function calls
    - [ ] method calls
    - [ ] closures
    - [ ] arrow functions
    - [ ] scope resolution operator
    - [ ] try/catch
    - [ ] numbers
    - [ ] numeric literal separator?
    - [ ] attributes
    - [ ] PHPDoc?
        - [ ] types
    - [ ] string escape sequences
    - [ ] punctuation
- [ ] folds seem to be working OK, but I don't use them very much, so it would be great if someone else could confirm that they are working as expected
- [ ] Figure out how to handle scopeing for builtin classes and functions
    - the TM grammar relies are large regexs for these, but these don't work w/ Atom's TS grammars
- [ ] TS includes the semicolons as it parses statements, but the TM grammar does not. This means that something like `use A;` will now have `meta.use.php` applied to the semicolon as well, whereas the TM grammar would stop that scope at `A`.
- [ ] need to scope `//` and `\` (as in `Foo\Bar`) as punctuation, but I'm having a hard time figuring out how to select them

### Alternate Designs

The main alternate I thought about was leaving the new specs in Coffeescript to match the existing specs. They would probably be alot easier to read in Coffeescript, as JS+prettier is turning 1 line of CS into 5-8+ lines of JS.

### Benefits

Finishes up a long-standing PR; speeds up PHP highlighting, folding, indenting, etc in Atom.

### Possible Drawbacks

Until https://github.com/atom/atom/pull/22130 is resolved, and `tree-sitter-php` is released w/ support for newer language features, the tree-sitter grammar will lack support for newer language features that are already supported by the existing TM grammar. This could be seen as a regression.

### Applicable Issues

#303


claytonrcarter:
Oh, hmm. Should I not be including the merge commit from `master` in this PR? What's the best practice for that. Technically, the only changes in this PR are to files 
```
grammars/tree-sitter-php.cson
specs/tree-sitter-spec.js
package.json
package-lock.json
```
All of the other changes are from master...

claytonrcarter:
Also, any assistance on debugging the window's builds would be much appreciated. Specs are passing on linux and macOS, but it looks like something is causing the tree-sitter grammar to not load on Windows because the specs appear to be using the existing TM grammar instead of the TS grammar.

Possibly related: https://github.com/atom/atom/issues/22675

KapitanOczywisty:
Best would be to rebase changes after current master's head. Just make sure you don't change commits from master. Easiest way which comes to mind: You can make patch from your commits (you can squash them) apply onto original `mb-tree-sitter-parser` and then make rebase.

And make PR to master.

About windows build: cleanup commits and I'll try this on my end, but you should not be concerned about that if this is atom's problem.

claytonrcarter:
OK thanks! I've rebased this branch (including original commits from #303) onto `master`, and updated this PR to target `master`. PLMK if I did anything wrong in that process.

B/c this is still draft, I would prefer to keep my commits unsquashed as I work, but I agree that they should be squashed before (if?) this gets merged.

KapitanOczywisty:
Looks good now.

KapitanOczywisty:
Seems like windows problems are caused by different node versions used for tree-sitter-php compilation. It will probably solve itself when that package will be bundled with atom itself, what will happen after update of package.json on master.

claytonrcarter:
I'm ported over as many of the existing tests as possible, and things are largely working OK, aside from the BCs mentioned in the description, many of which are out of our control at this point. Two things that *are* in our control, though are:
- figuring out how to scope the punctuation that I mentioned in the descriptions. I don't know why it's so tricky to scope separators in classnames and comments, etc, but it's really evading me.
- figuring a way to scope all of the standard library builtins. It would be great if the TS grammar could share the regexs used by the TM grammar, but Atom's TS support appears to limit how the regexs are used. eg it appears that multiline regexs w/ `(?xi)` are not allowed, and these are used extensively by the TM grammar.

KapitanOczywisty:
About punctuation problems: get a few most popular themes, get some php source codes like symfony, and check if it's looking ok. TM has more scopes that is ever used, so if not needed - move on.

> * figuring a way to scope all of the standard library builtins. It would be great if the TS grammar could share the regexs used by the TM grammar, but Atom's TS support appears to limit how the regexs are used. eg it appears that multiline regexs w/ `(?xi)` are not allowed, and these are used extensively by the TM grammar.

Function and constant names are out of date by now. I'm thinking about regenerating these from php source. For the first release could be probably skipped.

I'll try to get tree-sitter running on my machine to help with some grammar, but maybe you could get some help in `tree-sitter` or `tree-sitter-php` repositories.

grammars/tree-sitter-php.cson
```diff
@@ -0,0 +1,322 @@
...
+  'formal_parameters > "("': 'punctuation.definition.parameters.begin.bracket.round.php'
+  'formal_parameters > ")"': 'punctuation.definition.parameters.end.bracket.round.php'
+  'anonymous_function_use_clause > "("': 'punctuation.definition.parameters.begin.bracket.round.php'
+  'anonymous_function_use_clause > ")"': 'punctuation.definition.parameters.end.bracket.round.php'
+  'catch_clause > "("': 'punctuation.definition.parameters.begin.bracket.round.php'
+  'catch_clause > ")"': 'punctuation.definition.parameters.end.bracket.round.php'
+  'array_creation_expression > "("': 'punctuation.definition.array.begin.bracket.round.php'
+  'array_creation_expression > ")"': 'punctuation.definition.array.end.bracket.round.php'
+  'object_creation_expression > declaration_list > "{"': 'punctuation.definition.class.begin.bracket.curly.php'
+  'object_creation_expression > declaration_list > "}"': 'punctuation.definition.class.end.bracket.curly.php'
```
claytonrcarter, line 227: This matches the TM grammar, but feels like it's unnecessarily specific. For example, the [JS grammar](https://github.com/atom/language-javascript/blob/master/grammars/tree-sitter-javascript.cson#L164-L170) has all parens as just `punctuation.definition.parameters.(begin|end).bracket.round`. I would like to change all of these to just `punctuation.definition.(begin|end).bracket.round.php` and see how that affects themes.

KapitanOczywisty: Most of uses lists all of them: https://grep.app/search?q=punctuation%5C.definition%5C.%5Ba-z.%5D%2B%5C.bracket%5C.%5Ba-z.%5D%2B%5C.php&regexp=true

If is working, I'd personally leave this as is, but I don't think that change will break anything. Up to you.

KapitanOczywisty: We could also later make PR to replace this in both TS and TM, and gather feedback.

claytonrcarter: Ah, OK, makes sense. For now I've rearranged them so that all of the array/arguments/parameters are least grouped together. Thanks!

claytonrcarter:
Another quick update: this is not finished, but it has been my daily PHP driver for ~4 months now and -- as far as I can tell -- seems to work just fine. There are still a number of regressions between the TS grammar and the TM grammar, and most of these are noted in the PR description.

There are 2 *main* things that have changed since I last posted an update:
 - I got impatient waiting for some compatibility updates/downgrades and have started using a custom "fork" of `tree-sitter-php` that is really just `master` but w/ the parser generated by `tree-sitter` 0.17 instead of 0.20. This means it's using an ABI that's compatible w/ Atom's version of tree-sitter. I don't see a simpler way around this until [Atom updates it's version of tree-sitter](https://github.com/atom/atom/pull/23283)
 - Facing the difficulties of the `text.html.php`/`source.php` question, I changed the parser structure to match how the TM grammar does it: there are separate tree-sitter parsers for the "HTML" part of every PHP file (aka `tree-sitter-embedded-php` and another for the actual PHP parts (aka `tree-sitter-php`). The HTML parser is just a lightweight wrapper that injects the PHP parser. This allows us to properly (I think!, but ??) apply the `text.html.php` scope separately from the `source.php` scope, and closely matches how the TM grammar does it.
 - Not a big change, but this also includes `phpdoc` support. 

I think that these are mostly working correctly, but there are a number of differences from the TM grammar that are flagged in the `tree-sitter-spec.js` file. 

**NOTE/WARNING/WATCH OUT**  
In order to get this working, I created 3 separate dependencies in `package.js`:
 - `tree-sitter-embedded-php` - this is my TS parser for "html w/ embedded php", to mimic how the TM grammar works.  Code is at https://github.com/claytonrcarter/tree-sitter-embedded-php and on npm at https://www.npmjs.com/package/tree-sitter-embedded-php (2 weekly downloads! w00t!)
 - `tree-sitter-php-abc` - this is my "custom fork" of `tree-sitter-php` to support TS 0.17, and the code is hosted at https://github.com/claytonrcarter/tree-sitter-php/tree/atom-backport and on NPM at https://www.npmjs.com/package/tree-sitter-php-abc
 - `tree-sitter-phpdoc` - this is my TS parser for phpdoc, code is at https://github.com/claytonrcarter/tree-sitter-phpdoc and on npm at https://www.npmjs.com/package/tree-sitter-phpdoc

I know that this is not maintainable in the long run, but it felt like a decent way to get the ball rolling until something more sustainable can be figured out.

A few months have passed since I worked on this code, and I see now that there are a few updates to bring in from `tree-sitter-php` (not many) and from `language-php@0.48.1`. (As far as I can recall, this is up-to-date w/ `language-php@0.48`.)

@sadick254 @darangi Would greatly appreciate any input or guidance on this, if you can spare the time.

claytonrcarter:
It looks like CI is failing, but the tests are passing for me locally ... of course. 🙄 I'll try to look into it, but I won't have time to do so today.

claytonrcarter:
I've been chipping away at some issues w/ the embedding grammar the past few days. This isn't something that I've looked at much before, but I'm tinkering w/ having some add'l tests to confirm that the embedding tree-sitter grammar is up to snuff w/ the embedding TextMate grammar.

In particular, shebangs (eg `#!/usr/bin/env php`) currently aren't recognized or highlighted by the tree-sitter grammar. Locally, I have them recognized, but not yet highlighted. This is sort of interesting (at least, to me 😄) b/c the shebang needs to come before the opening php tag. W/o a special case for this, though, anything outside of php tags is considered HTML, and parsed/highlighted as such. In order for this to work w/i Atom's tree-sitter system, I think I'll need to work on an exception where the first line (and only the first line) can be considered "PHP" even if it's not w/i the start tag, but only if it's a shebang. We'll see...

mikehaertl:
@claytonrcarter I came here just by coincidence from the linked issue in the ts php parser. But I stumbled over hash bang handling in the js parser yesterday: https://github.com/tree-sitter/tree-sitter-javascript/blob/master/grammar.js#L92

Maybe it helps.

claytonrcarter:
Just rebased this to update it and squashed my ~70 commits down into a more reasonable handful. Though it's not without some issues, this is still my daily driver.

PR #439, opened 2021-08-11
Replace noreturn with never

KapitanOczywisty:
I haven't noticed that they changed actual name from `noreturn` to `never`.

https://php.watch/versions/8.1/never-return-type

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -690,8 +690,8 @@
             'name': 'storage.type.php'
           }
           {
-            'match': '\\b(noreturn)\\b',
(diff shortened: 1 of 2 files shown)


KapitanOczywisty:
8.1 is out now, can this be merged? @sadick254 ?

darangi:
Thanks for the contribution 🙇🏾  @KapitanOczywisty

Moriphete:
> I haven't noticed that they changed actual name from `noreturn` to `never`.
> 
> https://php.watch/versions/8.1/never-return-type

👽🔥👍🤘

Issue #449, opened 2022-02-05
When SQL used within a PHP string, interpolated PHP variables remain unstyled


keevan:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: <https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode>
    * Followed all applicable steps in the debugging guide: <https://flight-manual.atom.io/hacking-atom/sections/debugging/>
    * Checked the FAQs on the message board for common solutions: <https://discuss.atom.io/c/faq>
    * Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom>
    * Checked that there is not already an Atom package that provides the described functionality: <https://atom.io/packages>

### Description

Out of the box, VSCode seems to not encounter this particular issue with SQL highlighting within a PHP string. The highlighting seems to interpret the PHP variable as a string, when it probably should fallback to the language of the file / outer scope and check what sort of class it should have - in this particular case, it should be interpreted as a PHP variable.

Also, I might create another issue for the following, but not sure how easy it is to check for this problem:
- [ ] When using SQL inside a PHP string, it isn't styling joined strings (e.g. concat via `.`) and 
- [ ] When starting the string with a lowercase `select`, this does not trigger SQL highlighting. I could override the styles and transform them to lowercase, but this might actually be a project's styling convention, to not have the code be screaming at you.

### Steps to Reproduce

1. Create new file
2. Paste in the following:
```php
<?php
  $editors = '"'.implode('","', ['vim', 'atom', 'vscode']).'"';
  $results = database::select_one_column(
      "SELECT user.id
         FROM user
        WHERE user.healthy = true
          AND user.editor IN ($editors)"
  );
```

**Expected behavior:**

$editors should be highlighted/styled like a PHP variable

Screenshot from VSCode

![image](https://user-images.githubusercontent.com/9924643/152632605-275d8049-1ee8-41e7-b769-5a6c5dc6d58f.png)

**Actual behavior:**

$editors is styled like any other plain PHP string. 

Screenshot from Atom

![image](https://user-images.githubusercontent.com/9924643/152632619-2ff4736a-93b6-40de-a365-a2bb4d0ef4af.png)

**Reproduces how often:**

100% of the time

### Versions

Atom    : 1.58.0
Electron: 9.4.4
Chrome  : 83.0.4103.122
Node    : 12.14.1

[...]

Issue #87, opened 2015-06-29
Embedded SQL-Syntax only works for inline SQL


screeny05:
Used Atom Version: 1.0
Expected behaviour: Correct highlighting of SQL within PHP-Strings
Has this ever worked: No

Highlighting of SQL within strings is a very useful feature, but there's a little problem:

the language-php package only seems to highlight them if the beginning of the sql-statement lies on the same line as the opening of the string with quotation marks.

![image](https://cloud.githubusercontent.com/assets/765896/8409711/f5b93b6e-1e79-11e5-81fe-d733810563ad.png)
(Notice, that Lines 6-8 are all green)

zr9:
Atom uses textmate like grammar, if main principles of grammar is same then this one is impossible.
Textmate grammar limited to 1 rule per line, so it not possible to match few lines with 1 regex, here is info about that behavior  [Textmate Grammar](http://manual.macromates.com/en/language_grammars.html)

That problem require a multiline matching of text, because it is not possible otherwise distinguish SQL string from regular string basing only on **quote**.

Someone can fill bug/question about grammar to core devs?

thisispiers:
any updates? this issue also occurs with concatenated, embedded sql

zr9:
@thisispiers You probably talking about other one, could you post test example which code broke highlight? That one is related only to case when you not get correct highlight if SQL is not single line.

Arcanemagus:
This is the same issue as #52, the switch to SQL gets broken with any break in the string. (Including, apparently, a newline.)

50Wliu:
@Arcanemagus I'm going to keep this one separate just in case.

zr9:
@Arcanemagus @50Wliu 
In concept of bugs this is 2 different issues, #52 can be solved. When this one almost nope. Since tokenizer not support multiline capture

50Wliu:
> Since tokenizer not support multiline capture

Yes it does: just use `begin`/`end` instead of `match`.

zr9:
@50Wliu Yup in global it does. Problem so being and end is match only edges, and in that case there will be no different between SQL inside quotes and Str inside quotes. For detect this case there is need proper multiline regex with checks what is between begin and end. So for example is not possible to make expression(as simple example) like `/"\s*SELECT.*"/` because part `.*` not catch multiline.

[...]

PR #450, opened 2022-02-05
Fix some sql highlighting

keevan:
### Requirements

### Description of the Change

Combined screenshot of result

Before:

image

After (note `$editors` not being highlighted when SQL syntax is applied is a separate issue):

image

3 main changes to allow for SQL syntax highlighting in strings, in new scenarios.
1. Allows lowercased keywords (e.g. `"select * from users"`)

2. Query starting on next line of string opener (requires `--sql` on opening line). 

3. Query starting with a parenthesis will be allowed in the SQL begin check. (e.g. `"(select 1) union all (select 2)`"

### Alternate Designs
1. An alternative to this might be to repeat the same list of keywords in lowercase, so it prevents false positives such as the string "Select from the following:" from being incorrectly highlighted. This doesn't prevent the problem entirely though, and I'm probably still okay making a non-important string have weird colors than to not have this option for people who prefer code to - and I quote - "be screaming at them". 
2. Alternative options were discussed in this thread: https://github.com/atom/language-php/issues/87
Until tree-sitters become a reality, I think having a 'experimental / opt-in' option might make more people happy in the mean time. At least, I would be. An option is better than none / waiting imo. I wouldn't necessarily enforce it throughout the codebase though.
3. Heredocs were proposed but PHPCS might disallow such usage in the first place, and therefore could not be an option.

### Benefits
More syntax highlighting options for SQL in PHP strings, particularly ones that probably should have worked out of the box (1,2)

### Possible Drawbacks
False positives (1 proposed) - for the string examples mentioned above. 

Otherwise, if it's as the (alternative design 1), this would mean that sPonGebob notation will not be supported. (See https://sqlfum.pt/)

image

### Applicable Issues
https://github.com/atom/language-php/issues/87
https://github.com/atom/language-php/issues/449 (mentions but not its own issue)

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -2759,7 +2759,7 @@
     'match': '\\\\(?:\\\\(?:\\\\[\\\\\']?|[^\'])|.)'
     'name': 'constant.character.escape.php'
   'sql-string-double-quoted':
-    'begin': '"\\s*(?=(SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b)'
+    'begin': '(?i:"\\s*(?=(SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b))'
     'beginCaptures':
       '0':
         'name': 'punctuation.definition.string.begin.php'
@@ -2828,7 +2828,7 @@
       }
     ]
   'sql-string-single-quoted':
-    'begin': '\'\\s*(?=(SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b)'
+    'begin': '(?i:\'\\s*(?=(SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b))'
     'beginCaptures':
       '0':
         'name': 'punctuation.definition.string.begin.php'
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -2759,7 +2759,7 @@
     'match': '\\\\(?:\\\\(?:\\\\[\\\\\']?|[^\'])|.)'
     'name': 'constant.character.escape.php'
   'sql-string-double-quoted':
-    'begin': '(?i:"\\s*(?=(SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b))'
+    'begin': '(?i:"(?=\\s*\\(?\\s*(SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b))'
     'beginCaptures':
       '0':
         'name': 'punctuation.definition.string.begin.php'
@@ -2828,7 +2828,7 @@
       }
     ]
   'sql-string-single-quoted':
-    'begin': '(?i:\'\\s*(?=(SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b))'
+    'begin': '(?i:\'(?=\\s*\\(?\\s*(SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b))'
     'beginCaptures':
       '0':
         'name': 'punctuation.definition.string.begin.php'
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -2759,7 +2759,7 @@
     'match': '\\\\(?:\\\\(?:\\\\[\\\\\']?|[^\'])|.)'
     'name': 'constant.character.escape.php'
   'sql-string-double-quoted':
-    'begin': '(?i:"(?=\\s*\\(?\\s*(--\\s*SQL|SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b))'
+    'begin': '(?i:"(?=\\s*\\(?\\s*(--\\s+SQL|SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b))'
     'beginCaptures':
       '0':
         'name': 'punctuation.definition.string.begin.php'
@@ -2828,7 +2828,7 @@
       }
     ]
   'sql-string-single-quoted':
-    'begin': '(?i:\'(?=\\s*\\(?\\s*(--\\s*SQL|SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b))'
+    'begin': '(?i:\'(?=\\s*\\(?\\s*(--\\s+SQL|SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b))'
     'beginCaptures':
       '0':
         'name': 'punctuation.definition.string.begin.php'
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -2759,7 +2759,7 @@
     'match': '\\\\(?:\\\\(?:\\\\[\\\\\']?|[^\'])|.)'
     'name': 'constant.character.escape.php'
   'sql-string-double-quoted':
-    'begin': '(?i:"(?=\\s*\\(?\\s*(SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b))'
+    'begin': '(?i:"(?=\\s*\\(?\\s*(--\\s*SQL|SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b))'
     'beginCaptures':
       '0':
         'name': 'punctuation.definition.string.begin.php'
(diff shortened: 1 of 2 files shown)


keevan:
- [ ] Should fix up `--sql` matching to be at least `-- sql` instead, because `--sql` is not a valid comment and syntax highlighting should not incorrectly kick off earlier than intended.

grammars/php.cson
```diff
@@ -2759,7 +2759,7 @@
     'match': '\\\\(?:\\\\(?:\\\\[\\\\\']?|[^\'])|.)'
     'name': 'constant.character.escape.php'
   'sql-string-double-quoted':
-    'begin': '"\\s*(?=(SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b)'
+    'begin': '(?i:"(?=\\s*\\(?\\s*(--\\s+SQL|SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b))'
```
KapitanOczywisty, line 2762:
```suggestion
    'begin': '(?i)"(?=\\s*\\(?\\s*(SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b)'
```

grammars/php.cson
```diff
@@ -2828,7 +2828,7 @@
       }
     ]
   'sql-string-single-quoted':
-    'begin': '\'\\s*(?=(SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b)'
+    'begin': '(?i:\'(?=\\s*\\(?\\s*(--\\s+SQL|SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b))'
```
KapitanOczywisty, line 2831:
```suggestion
    'begin': '(?i)\'(?=\\s*\\(?\\s*(SELECT|INSERT|UPDATE|DELETE|CREATE|REPLACE|ALTER|AND|WITH)\\b)'
```

KapitanOczywisty (changes requested):
> 1\. Allows lowercased keywords (e.g. `"select * from users"`)
> 3\. Query starting with a parenthesis will be allowed in the SQL begin check. (e.g. `"(select 1) union all (select 2)`"

These are ok for me

> 2\. Query starting on next line of string opener (requires `--sql` on opening line).

However I don't see any point for that one. There Is Heredocs for such cases, and `-- sql` seems to not be used at all in that context in the wild: https://grep.app/search?q=%5B%27%22%5D--%5Cs%2Bsql&regexp=true&filter[lang][0]=PHP

Also please add some tests for new features.

KapitanOczywisty:
> 1. An alternative to this might be to repeat the same list of keywords in lowercase, so it prevents false positives such as the string "Select from the following:" from being incorrectly highlighted.

This is valid concern, but `Select` is also used for sql ([examples](https://grep.app/search?q=%5B%22%27%5DSelect%20%5C%2A&regexp=true&case=true&filter[lang][0]=PHP)). I'm not sure if it's worth adding other  variants than uppercased, since it can break a lot of code. But this is something to be discussed.

keevan:
> Rule was created when heredocs weren't able to handle indentations, this is not the case now. By grep.app [`Squiz.PHP.Heredoc.NotAllowed` was disabled more often](https://grep.app/search?q=Squiz.PHP.Heredoc.NotAllowed&filter%5Bpath.pattern%5D%5B0%5D=phpcs.xml&filter%5Blang%5D%5B0%5D=XML) than [`-- sql` was ever used](https://grep.app/search?q=%5B%27%22%5D--%5Cs%2Bsql&regexp=true&filter%5Blang%5D%5B0%5D=PHP).

Fair point. It seems the indent-able heredocs was introduced in php 7.3. Also yeah I wouldn't be suprised if  `-- sql` was ever used anywhere, since it started from a need here: https://github.com/atom/language-php/issues/87#issuecomment-238056336

I still personally prefer the option to trigger highlighting with `-- sql` because the effort to add it to existing query blocks is simpler than converting all the bits to heredocs. But that's what forks are for :smile: (Though I wish there were an option to toggle this on / off in settings)

Also thanks for the feedback. I'll try to add some tests for the new stuff when I get the chance.

keevan:
@KapitanOczywisty I've applied your suggestions and added some tests based on tests that were already there. Let me know if you prefer I split it out to a separate `it` block, though I feel like it made more sense where I've added it.

Ran specs locally - seems okay.

image

adjenks:
Something that really bugs me is that the CTE keyword `WITH` doesn't trigger syntax highlighting. Would that be fixed if I added the `--sql`?

keevan:
> Something that really bugs me is that the CTE keyword `WITH` doesn't trigger syntax highlighting. Would that be fixed if I added the `--sql`?

Ah, the `--sql` keyword was something I thought about adding, but has been removed as @KapitanOczywisty did not want to introduce new custom syntax to trigger the syntax highlighting. 

I tried locally, and it seems like it is highlighting as expected:

image

The only time it doesn't is if the string starts with a `;`. @adjenks, Do you have an example of when you'd expect it to start highlighting? Perhaps it might be simple enough to include as part of this PR as well to hopefully get it over the line.

adjenks:
@keevan It doesn't exist in phpDocumentor, but I always thought that putting the language in a [doc-block](https://en.wikipedia.org/wiki/Docblock) made the most sense. It's common format for documenting code.

Something like...
/** @lang sql */
 $query = 'SELECT 1;';

There's no lang tag, but, here is the var tag as an example: 
https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/var.html#examples

I don't think it should be inside the string itself because then you add some overhead to the query being sent to the server, albeit not much.

KapitanOczywisty:
> The only time it doesn't is if the string starts with a `;`.

What is the purpose of `;` before the query? I don't think it is functional unless you are creating multi-query - which is not the best idea in php.

> It doesn't exist in phpDocumentor, but I always thought that putting the language in a [doc-block](https://en.wikipedia.org/wiki/Docblock) made the most sense. It's common format for documenting code.

It won't work for TextMate (or even Tree-sitter) syntax highlighting, it would be way too much added complexity, if even possible.

> I don't think it should be inside the string itself because then you add some overhead to the query being sent to the server, albeit not much.

This is over-optimization, striping comments takes nanoseconds. Updating table indexes can have real impact on performance, not removing comments. Measure, optimize and measure again.

keevan:
> What is the purpose of ; before the query?

I don't use it myself, I only observed examples online when searching for examples for @adjenks where highlighting did not apply. 

With that said, if there's nothing left that needs to be changed, I'd like these changes to be reviewed and merged in if possible :smiley:

KapitanOczywisty:
> With that said, if there's nothing left that needs to be changed, I'd like these changes to be reviewed and merged in if possible 😃

This will take a while, also with possible false positives from case insensitive keywords, I'd like to see the input from the community.

crvllws (approved):

crvllws (approved):
Thanks

crvllws (approved):

crvllws (approved):

crvllws (approved):

crvllws (approved):

crvllws (approved):

crvllws (approved):

KapitanOczywisty:
@crvllws Stop spamming, you're not helping...

crvllws (approved):

adjenks:
> I'd like to see the input from the community.

I still think that all syntax highlighters should offer a way of indicating the language of a string by writing it in a preceding comment, much like how you often can in markdown by writing the language after three back-ticks.

Markdown:
> \```js
> console.log('hello');
> \```
Result:
```js
console.log('hello');
```

Perhaps we could have a vote on anything that you want input on?

We could vote on manual language selections here using reactions:

(React 😄) Option 1 -  Inside the text itself:
```PHP
$query = '--sql
SELECT 1';
```

(React 🎉) Option 2 - Inside a preceding docblock:
```PHP
/** @lang sql */
$query = 'SELECT 1';
```

(React 🚀) Option 3 - Inside a preceding non-docblock comment:
```PHP
//L:sql
$query = 'SELECT 1';
```

(React 😕) Option 4 - None, only automatic language detection:
```PHP
$query = 'SELECT 1';
```

I think auto-detection is fine, but if you can just manually select I think it would solve some problems and improve documentation and processing in some cases.

KapitanOczywisty:
@adjenks First of all atom is dead, so there is no point discussing this PR any further. If you're using vscode there will be fork by me or somebody else and there you can continue - with vscode in mind.

Second thing options 2 and 3 are not viable to implement, textmate grammar doesn't have "memory" to keep track of language labels and use them later. Option 1 is non-standard and not popular (invented in this very PR), this is not something which should be added for hundreds of thousands of users.

And finally there is heredoc `<<<SQL` which is used, popular and since php 7.3 properly formattable.

There was somwhere issue to improve semantic tokens in vscode which would allow to support stuff like `/** @lang sql */` without major overhead, however I don't think it is on the radar right now.

adjenks:
@KapitanOczywisty Ah okay. Thank you for all the info.

I was interested in highlighting for both applications since I was using both when I joined the discussion. Someone had given me the impression that VSCode shared this library. Feel free to point me to the correct library for VSCode if one has been made.

That's unfortunate about the unviability. The semantic token issue you mentioned would be nice. I'll see if I can find it.

KapitanOczywisty:
> Someone had given me the impression that VSCode shared this library.

It is still using it, but nobody is merging anything and repository will be archived in mid december.

> Feel free to point me to the correct library for VSCode if one has been made.

Not yet, but we're thinking about how to proceed. I'll leave note about the new repository in the issues when time comes.

PR #462, opened 2022-09-20
Allow named arguments in constructors

BladeMF:
### Requirements

### Description of the Change

Changes "instantiation" pattern to allow for named parameters when instantiating objects.

### Alternate Designs

I didn't find any other alternatives than changing the "instantiation" pattern.

### Benefits

The names of the parameters will be tokenized in constructors.

### Possible Drawbacks

Can't see any.

### Applicable Issues

No applicable issues that I know of.

### DIFF
--- a/grammars/php.cson
+++ b/grammars/php.cson
@@ -2316,36 +2316,59 @@
       }
     ]
   'instantiation':
-    'begin': '(?i)(new)\\s+(?!class\\b)([a-z0-9_\\x{7f}-\\x{10ffff}\\\\]+)\\s*(\\()'
-    'beginCaptures':
-      '1':
-        'name': 'keyword.other.new.php'
-      '2':
-        'patterns': [
-          {
-            'match': '(?i)(parent|static|self)(?![a-z0-9_\\x{7f}-\\x{10ffff}])'
-            'name': 'storage.type.php'
-          }
-          {
-            'include': '#class-name'
-          }
-          {
-            'include': '#variable-name'
-          }
-        ]
-      '3':
-        'name': 'punctuation.definition.arguments.begin.bracket.round.php'
-    'end': '\\)'
-    'endCaptures':
-      '0':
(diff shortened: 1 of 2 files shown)


BladeMF:
I have no idea why this test fails or how to fix it. Will appreciate help if there is something I need to do.

KapitanOczywisty:
`new FooBar;` is not supported after this change.

Also for anonymous class, the argument list (with parenthesis) is marked as `meta.function-call.php` and I'd like to keep this. `meta.instantiation.php` will be messy to add for anonymous classes so for consistency sake I'd not use it.

Lookup tokens for:
```php
new class(bar:312) {
  public function __construct($bar=0){}
};
```

KapitanOczywisty:
> Why is there no test for `new FooBar;`?

Base for this repo was adapted from different project and tests were mostly added for changes.

> Should I create one?

Yes.

> Otherwise I don't mind changing `meta.instantiation.php` for the other one.

👍 I'd keep only part with parentheses in `meta.function-call.php`.

> BTW what's wrong with the tests?

Someone changed dependency without updating this repository, I'll take a look at this. For now just ignore errors and test locally.

BladeMF:
Thank you for helping.

BladeMF:
> I'd keep only part with parentheses in `meta.function-call.php`.

Do you mean the part **_between_** the parentheses, or the whole expression in the **_variant_** that **_has_** parentheses? In functions the whole function call is marked as `meta.function-call.php`, along with the parentheses and function name.

BladeMF:
Done. We have these cases as tests now:
```php
$a = new $b;
$a = new className1;
$a = new $b();
$a = new className1(5, named: 4);
```

BladeMF:
While I was at it I added tests for:
```php
$a = new self;
$a = new self();
$a = new static;
$a = new static();
$a = new parent;
$a = new parent();
```

DiqahLer (approved):

PR #463, opened 2022-11-10
SQL string with \n \r or \t prefixed

LucianoHanna:
### Requirements

### Description of the Change

Identify SQL string with \n, \r or \t prefixed.

### Benefits

Will identify SQL string with \n, \r or \t prefixed as in the following example:

```php
DB::select("
    SELECT 1
");
```

### Possible Drawbacks

None

### Applicable Issues

None open


KapitanOczywisty:
`\s` includes all `\n`, `\r` and `\t`, the problem is that grammar interprets one line at the time - so this won't work.

